PR #84, opened 2020-02-02
Add toast

juliushaertl:
Basically a copy of https://github.com/nextcloud/server/blob/master/core/src/OCP/toast.js so we can remove it from the server at some point.

### DIFF
--- a/lib/toast.ts
+++ b/lib/toast.ts
@@ -17,28 +17,28 @@ export interface ToastOptions {
 	/**
 	 * Defines the timeout after which the toast is closed. Set to 0 to have a persistent toast.
 	 */
-	timeout: number;
+	timeout?: number;
 	/**
 	 * Set to true to allow HTML content inside of the toast text
 	 * @default false
 	 */
-	isHTML: Boolean;
+	isHTML?: Boolean;
 	/**
 	 * Set a type of {ToastType} to style the modal
 	 */
-	type: ToastType|undefined;
+	type?: ToastType|undefined;
 	/**
 	 * Provide a function that is called after the toast is shown
 	 */
-	callback: Function,
+	onShow?: Function,
 	/**
 	 * Provide a function that is called when the toast is clicked
 	 */
-	onClick: Function,
+	onClick?: Function,
 	/**
 	 * Make the toast closable
 	 */
-	close: Boolean;
+	close?: Boolean;
 }
 
 /**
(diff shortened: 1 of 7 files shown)


lib/toast.ts
```diff
@@ -0,0 +1,125 @@
...
+}
+
+/**
+ * Show a toast message
+ * 
+ * @param text Message to be shown in the toast, any HTML is removed by default
+ * @param options
+ */
+export function showMessage(text: string, options: ToastOptions): Toast {
+	options = options || {}
```
skjnldsv: can't you provide a default in the function declaration directly?

juliushaertl: Ah we can actually get rid of the line since the defaults are assigned in the next one. Fixed.

lib/toast.ts
```diff
@@ -0,0 +1,125 @@
...
+	 */
+	isHTML: Boolean;
+	/**
+	 * Set a type of {ToastType} to style the modal
+	 */
+	type: ToastType|undefined;
+	/**
+	 * Provide a function that is called after the toast is shown
+	 */
+	callback: Function,
```
ChristophWurst:
```suggestion
	onShow: Function,
```

lib/toast.ts
```diff
@@ -0,0 +1,125 @@
...
+	close: Boolean;
+}
+
+/**
+ * Show a toast message
+ *
+ * @param text Message to be shown in the toast, any HTML is removed by default
+ * @param options
+ */
+export function showMessage(text: string, options: ToastOptions): Toast {
```
ChristophWurst:
```suggestion
export function showMessage(text: string, options?: ToastOptions): Toast {
```

Unless they are mandatory

lib/toast.ts
```diff
@@ -0,0 +1,125 @@
...
+	static readonly INFO = 'toast-info';
+	static readonly SUCCESS = 'toast-success';
+	static readonly PERMANENT = 'toast-error';
+}
+
+export interface ToastOptions {
+	/**
+	 * Defines the timeout after which the toast is closed. Set to 0 to have a persistent toast.
+	 */
+	timeout: number;
```
ChristophWurst:
```suggestion
	timeout?: number;
```

Same for the other props. Unless they are mandatory to specify, we should make their types nullable

ChristophWurst:
> so we can remove it from the server at some point.

We can remove it from server as soon as we ship it here :)

lib/toast.ts
```diff
@@ -0,0 +1,125 @@
+import Toastify from 'toastify-js'
+
+declare interface Toast {
```
juliushaertl: @ChristophWurst Not sure about this since this is mainly a duplicate definition as it is already part of `toastify.d.td` but without having it here, I found no way to make sure it is also included in the documentation.

ChristophWurst: Add `--includeDeclarations --excludeExternals` to the `build:doc` script. That seems to do the trick :)

juliushaertl: Thanks, worked like a charm.

ChristophWurst:
> We can remove it from server as soon as we ship it here :)

https://github.com/nextcloud/server/pull/19346 :rocket:

PR #86, opened 2020-02-22
Toast: fix timeout and callback

korelstar:
This PR contains two fixes:

## Timeout

The documentation for persistent toasts was wrong. Instead of using `timeout=0` you must use a negative number, e.g. `timeout=-1`. See also https://github.com/apvarun/toastify-js/pull/26 .

## Callback

Toastify-js has a parameter `callback`. This callback function is called when removing the Toast from DOM, see https://github.com/apvarun/toastify-js/blob/1.6.1/src/toastify.js#L233-L245 . Unfortunately, in @nextcloud/dialogs, this parameter was named `onShow` (see https://github.com/nextcloud/nextcloud-dialogs/blob/v1.1.0/lib/toast.ts#L66), which is very misleading. Therefore, this PR renames the parameter and updates the docs.

### DIFF
--- a/lib/toast.ts
+++ b/lib/toast.ts
@@ -23,9 +23,9 @@ export interface ToastOptions {
 	 */
 	type?: ToastType|undefined;
 	/**
-	 * Provide a function that is called after the toast is shown
+	 * Provide a function that is called after the toast is removed
 	 */
-	onShow?: Function,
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Thanks :) Guess this would make the next version a 2.x then :)

PR #106, opened 2020-03-23
Add Undo method & translations

skjnldsv:
- [x] Toastify upstream https://github.com/apvarun/toastify-js/pull/54
- [x] Cleanup
- [x] Undo method
- [x] Translate

image


lib/toastify.d.ts
```diff
@@ -11,7 +11,8 @@ declare interface ToastElement {
 declare module 'toastify-js' {
 
     interface ToastifyOptions {
-        text: string,
+        text?: string,
```
skjnldsv: How can I make sure one of the two is provided?
Not optional :)

styles/close.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" height="16" width="16">
```
skjnldsv: This require us (devs) to properly setup inline base64 webpack config as we're importing the scss from our apps and not when building this lib

lib/toast.ts
```diff
@@ -82,7 +115,7 @@ export function showMessage(text: string, options?: ToastOptions): Toast {
 		selector: options.selector,
 		position: 'right',
 		backgroundColor: '',
-		className: 'toast ' + classes,
+		className: 'dialogs ' + classes,
```
skjnldsv: This completely removes compatibility with server's css. It will force devs to import the scss I guess?

Not fan, I would rather import the scss automatically just when we import this app.
ping @nextcloud/javascript

styles/toast.scss
```diff
@@ -11,38 +35,60 @@
...
+	&.toast-warning {
+		border-left: 3px solid var(--color-warning);
+	}
+
+	&.toast-success {
+		border-left: 3px solid var(--color-success);
+	}
+
+	&.toast-undo {
+		border-left: 3px solid var(--color-success);
```
skjnldsv: Success? Info?

lib/toast.ts
```diff
@@ -127,3 +160,35 @@ export function showInfo(text: string, options?: ToastOptions): Toast {
...
+	const undoButton = document.createElement('span')
+	undoButton.classList.add('toast-undo-button')
+	undoButton.innerText = 'UNDO'
+	undoContent.innerText = text
+	undoContent.appendChild(undoButton)
+
+	if (options.onUndo instanceof Function) {
+		undoButton.addEventListener('click', function(event) {
+			event.stopPropagation();
+			options.onUndo(event);
```
skjnldsv: Could be nice if we could close the toast directly on click. Any ideas how?

lib/toastify.d.ts
ChristophWurst: you can't, really. not even with enums in ts.

how about a new exported function?

lib/toast.ts
ChristophWurst: > Not fan, I would rather import the scss automatically just when we import this app.
> ping @nextcloud/javascript

Won't work, I think. You need the css/style loader in the app. And when this lib exports the styles it's just a js module. Thus no styles would be loaded unless inline, right?

styles/toast.scss
ChristophWurst: how "clean" is this? is there fallback for when the server var is not set?

lib/toast.ts
```diff
@@ -127,3 +160,35 @@ export function showInfo(text: string, options?: ToastOptions): Toast {
...
+		timeout: 10,
+		// remove close button
+		close: false
+	})
+
+	// Generate undo layout
+	const undoContent = document.createElement('span')
+	const undoButton = document.createElement('span')
+	undoButton.classList.add('toast-undo-button')
+	undoButton.innerText = 'UNDO'
```
korelstar: I think it would be better to generate upper case with style sheets (if this is really needed) - this would ease translations.

Maybe you can also add an option for the button label? In some cases it will be helpful to be able to be more concrete (especially for German translation: "Wiederherstellen" vs. "Rückgängig" after deleting something).

skjnldsv: Yes, this is missing translations :)
I will 100% do that with css :)

styles/toast.scss
skjnldsv: Nope, none as far as I know.
Worst case scenario is no border colour, so not an issue I guess :)

lib/toast.ts
skjnldsv: Well, plyr recently removed the manual import of their scss for example. SO it might be doable :)
Wouldn't any script be able to inject their styling anyway?

lib/toastify.d.ts
skjnldsv: > how about a new exported function?

How would that work?

lib/toast.ts
```diff
@@ -127,3 +160,35 @@ export function showInfo(text: string, options?: ToastOptions): Toast {
...
+	options = Object.assign(options || {}, {
+		// force 10 seconds of timeout
+		timeout: 10,
+		// remove close button
+		close: false
+	})
+
+	// Generate undo layout
+	const undoContent = document.createElement('span')
+	const undoButton = document.createElement('span')
```
korelstar: Should this be a button and look like a button?

skjnldsv: I'll deal with the styling after :)

lib/toast.ts
korelstar: you can call `hideToast()` on the result of `showMessage`

gary-kim: The returned `toast` has a function called `hideToast`. You could replace 
```javascript
return showMessage(undoContent, { ...options, type: ToastType.UNDO })
```
with
```javascript
const toast = showMessage(undoContent, { ...options, type: ToastType.UNDO })
return toast

function onClick(event) {
    event.stopPropagation()
    options.onUndo(event)
    toast.hideToast()
}
```
and make `onClick` be the `click` event handler?

skjnldsv: No, otherwise the whole notification triggers the undo :)
This is not what we want here :hugs:

gary-kim: Huh, I thought GitHub would update new messages without reload. Didn't see @korelstar 's message while writing.

skjnldsv: > EDIT: Also, I meant undoButton's addEventListener('click', onClick), which should only work on the undo button unless I'm misunderstanding what you mean.



Ah right. Well, I rather not prevent any default implementation. If devs want to use onUndo, onClick and onRemove at the time, so be it :)

gary-kim: Oh, that's not what I meant. I just named it `onClick` randomly. I didn't meant that should be the `onClick` method. Sorry for the misunderstanding.

skjnldsv: Ahah, sorry :p

skjnldsv: Done

lib/toast.ts
```diff
@@ -165,9 +160,17 @@ export function showSuccess(text: string, options?: ToastOptions): Toast {
  * Show a toast message with undo styling
  *
  * @param text Message to be shown in the toast, any HTML is removed by default
+ * @param onUndo Function that is called when the undo button is clicked
  * @param options
  */
-export function showUndo(text: string, options?: ToastOptions): Toast {
+export function showUndo(text: string, onUndo: Function, options: ToastOptions): Toast {
```
skjnldsv: Since there is no point of having an undo without the proper method, I set it as mandatory.
Anyone against?

lib/toast.ts
```diff
@@ -127,3 +155,48 @@ export function showInfo(text: string, options?: ToastOptions): Toast {
...
 }
+
+/**
+ * Show a toast message with undo styling
+ *
+ * @param text Message to be shown in the toast, any HTML is removed by default
+ * @param onUndo Function that is called when the undo button is clicked
+ * @param options
+ */
+export function showUndo(text: string, onUndo: Function, options: ToastOptions): Toast {
```
skjnldsv: Since there is no point of having an undo without the proper method, I set it as mandatory.
Anyone against?

juliushaertl: Sure, but options are still optional :wink:

skjnldsv: Oups :grimacing:

lib/toastify.d.ts
ChristophWurst: okay I though these options are for *our* exported functions, but this just types the toastify-js lib. you can't make sure either of the two props are set. just don't make any mistakes :P

skjnldsv: Ok then :)
Thanks!

lib/toast.ts
ChristophWurst: Anything's possible :man_shrugging:

jancborchardt:
- [ ] Please include the name of the deleted element, like: `Deleted form "Product survey"`
- [ ] Undo should not be all-caps but "Undo", we never all-caps cause it is more difficult to read.
- [ ] The "Undo" needs to look clickable – ideally by being a button. Less ideally by being bold and underlined and primary-color like a link.

gary-kim:
>     * Please include the name of the deleted element, like: `Deleted form "Product survey"`

I think that's up to whatever app calls `showUndo` to set the appropriate text.

skjnldsv:
> I think that's up to whatever app calls `showUndo` to set the appropriate text.

Yep :)

jancborchardt:
> > * [ ]  Please include the name of the deleted element, like: `Deleted form "Product survey"`
> 
> I think that's up to whatever app calls `showUndo` to set the appropriate text.

If we shorten the text to "Deleted `element`" (like "Deleted Product survey" or "Deleted filename.txt"), then we can have this generic in the component, no?

Cause it would be much nicer if we standardize the wording in the component as well and don’t have to make sure every app does it properly. That way it would work for "Deleted", "Moved", "Renamed" etc., which can be used for files, forms, calendar, contacts, … and we only have to do the logic and translation once, in the component.

skjnldsv:
> Cause it would be much nicer if we standardize the wording in the component as well and don’t have to make sure every app does it properly. That way it would work for "Deleted", "Moved", "Renamed" etc., which can be used for files, forms, calendar, contacts, … and we only have to do the logic and translation once, in the component.

Then we would have to have one method per type of Undo?
undoDeleted, undoMoved, undoRenamed? Or even if it's not from the method name itself, it will have to be passed as an argument somewhere.

I'm all in for standardisation, but this is too much here :)

jancborchardt:
> > Cause it would be much nicer if we standardize the wording in the component as well and don’t have to make sure every app does it properly. That way it would work for "Deleted", "Moved", "Renamed" etc., which can be used for files, forms, calendar, contacts, … and we only have to do the logic and translation once, in the component.
> 
> Then we would have to have one method per type of Undo?
> undoDeleted, undoMoved, undoRenamed? Or even if it's not from the method name itself, it will have to be passed as an argument somewhere.
> 
> I'm all in for standardisation, but this is too much here :)

I’d rather we have the work here centrally, than have to check it in every single place. The action and UI is standardized and so should the wording and how it is displayed. It will ease the work for _all_ developers, and make it consistent for _all_ users.

If we don’t do that, at least make sure the wording recommendations are in the documentation (which hopefully all devs will read).

skjnldsv:
Toast library have been updated with my changes!
We can move this on!! :rocket:

skjnldsv:
All good! Please review everyone :tada:

skjnldsv:
So, it seems something broke, I don't understand why. 
I pushed whatever I had left, @juliushaertl maybe you could have a quick look? Since you had more experience with rollup :'(

Or anyone, It seems it's a config issue where the imports are weird misbehaving. I don't understand why

l10n/fr.pot
```diff
@@ -0,0 +1,20 @@
...
+"Last-Translator: Julien Veyssier <eneiluj@gmx.fr>, 2020\n"
+"Language-Team: French (https://www.transifex.com/nextcloud/teams/64236/fr/)\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Language: fr\n"
+"Plural-Forms: nplurals=2; plural=(n > 1);\n"
+
+#: src/components/MultiselectTags/MultiselectTags.vue:170
+#: lib/toast.ts:187
+msgid "Undo"
+msgstr "Annuler"
```
skjnldsv, line 20: Dummy file, will get overrided, it's fine

lib/toast.ts
```diff
@@ -1,16 +1,40 @@
...
 	static readonly INFO = 'toast-info';
 	static readonly SUCCESS = 'toast-success';
 	static readonly PERMANENT = 'toast-error';
+	static readonly UNDO = 'toast-undo';
 }
 
 export interface ToastOptions {
 	/**
-	 * Defines the timeout after which the toast is closed. Set to 0 to have a persistent toast.
+	 * Defines the timeout after which the toast is closed. Set to -1 to have a persistent toast.
```
nickvergessen, line 37: As per #107 and #111 this is not intended.

PR #107, opened 2020-04-07
Allow to set -1 as duration so toasts don't expire

nickvergessen:
Required for talk so the "signaling server error" message can be kept until it works again.

### DIFF
--- a/lib/toast.ts
+++ b/lib/toast.ts
@@ -71,10 +71,18 @@ export function showMessage(text: string, options?: ToastOptions): Toast {
 	}
 	let classes = options.type ?? ''
 
+	let duration = null
+	if (options.timeout) {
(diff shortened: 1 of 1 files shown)


ChristophWurst:
https://github.com/apvarun/toastify-js#api says 0 is used for permanent toasts. Why would you use -1?

nickvergessen:
Bildschirmfoto von 2020-04-07 10-59-53


PHPStorm got me confused. 0 seems to work.

ChristophWurst:
Could you please update https://github.com/nextcloud/nextcloud-dialogs/blob/88fd103076ac728aaa0b967fa1b86b5111192b27/lib/toast.ts#L12-L15 with the new knowledge? :)

PR #111, opened 2020-04-15
Adjust docs for persistent toast

nickvergessen:
See #107

### DIFF
--- a/lib/toast.ts
+++ b/lib/toast.ts
@@ -10,7 +10,7 @@ class ToastType {
 
 export interface ToastOptions {
 	/**
-	 * Defines the timeout after which the toast is closed. Set to -1 to have a persistent toast.
+	 * Defines the timeout after which the toast is closed. Set to 0 to have a persistent toast.
 	 */
 	timeout?: number
 


korelstar:
Sorry, but this doesn't work! Due to https://github.com/nextcloud/nextcloud-dialogs/blob/aab06d3e41fa72dfbcc69ab3a3c2a0cdcef765b2/lib/toast.ts#L77 , `Toastify` is called with option `duration=null` which results in using the default value of 3000 (see https://github.com/apvarun/toastify-js#api).

Therefore, you must use `timeout = -1` for a permanent toast or fix https://github.com/nextcloud/nextcloud-dialogs/blob/aab06d3e41fa72dfbcc69ab3a3c2a0cdcef765b2/lib/toast.ts#L77

korelstar:
@nickvergessen @ChristophWurst  Did you test this?

ChristophWurst:
> @nickvergessen @ChristophWurst Did you test this?

Nope :grin:

nickvergessen:
So I was right in the beginning with #107 which fixed it.
Afterwards I though I'm too stupid

ChristophWurst:
Just that the lib says you should use 0. Let's stick to that.

PR #118, opened 2020-05-23
Fix Dialogs not sticky

jotoeri:
Fixes nextcloud/forms/issues/398


ChristophWurst:
let me know if you need a release for this

jotoeri:
Yes, would be cool, if there is one before forms-release. So no need to hurry, but i think within the next 3-4 weeks this should be the case...

ChristophWurst:
done :)

jotoeri:
Nice, thank you! :) 👍

PR #166, opened 2020-08-19
Bundle toastify and transpile it

skjnldsv:
Otherwise it break our apps.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,8 +26,9 @@
   },
   "dependencies": {
     "@nextcloud/l10n": "^1.3.0",
-    "@nextcloud/typings": "latest",
-    "core-js": "^3.6.4"
+    "@nextcloud/typings": "^0.2.2",
+    "core-js": "^3.6.4",
+    "toastify-js": "^1.9.1"
   },
   "devDependencies": {
     "@babel/cli": "^7.8.4",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -26,8 +26,10 @@
   },
   "dependencies": {
     "@nextcloud/l10n": "^1.3.0",
-    "@nextcloud/typings": "latest",
-    "core-js": "^3.6.4"
+    "@nextcloud/typings": "^0.2.2",
+    "babel-loader-exclude-node-modules-except": "^1.0.3",
```
juliushaertl, line 30: Should be a dev dependency I'd say

juliushaertl (approved):
Otherwise :+1:

package.json
skjnldsv: oupsy!

korelstar:
Do I understand correctly that toastify is now included in the npm artifact and apps don't need to install toastify on their own?

If yes: should toastify (and typings and l10n) be a dev dependency? Only then, apps won't have to install those packages ...

skjnldsv:
> If yes: should toastify (and typings and l10n) be a dev dependency? Only then, apps won't have to install those packages ...

Hum, good call!

PR #178, opened 2020-08-27
Fix permanent displaying

nickvergessen:
Regression in conflict resolution from #106

See #107  and #111 for discussion.

### DIFF
--- a/lib/toast.ts
+++ b/lib/toast.ts
@@ -102,9 +102,14 @@ export function showMessage(data: string|Node, options?: ToastOptions): Toast {
 
 	const isNode = data instanceof Node
 
+	let timeout = null
+	if (options.timeout) {
+		timeout = options.timeout === -1 ? -1 : options.timeout * 1000
+	}
+
 	const toast = Toastify({
 		[!isNode ? 'text' : 'node']: data,
-		duration: options.timeout ? options.timeout * 1000 : null,
+		duration: timeout,
 		callback: options.onRemove,
 		onClick: options.onClick,
 		close: options.close,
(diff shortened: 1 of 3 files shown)


skjnldsv (commented):
I don't get it, Toastify requires -1 for permanent
https://github.com/apvarun/toastify-js/blob/master/README.md#api

nickvergessen:
https://github.com/apvarun/toastify-js/commit/4e675736f1154d42c49c133d9f85a2ae49db2ad3

okay this changed with 1.8.0 then let me update the readme instead.

skjnldsv (approved):
sure

lib/toast.ts
```diff
@@ -175,7 +180,7 @@ export function showUndo(text: string, onUndo: Function, options?: ToastOptions)
 
 	options = Object.assign(options || {}, {
 		// force 10 seconds of timeout
-		timeout: 1000,
+		timeout: 10000,
```
nickvergessen, line 183: Because of this I would actually do a 2.0.1 release.

skjnldsv: Right!

skjnldsv:
So, this actually broke Undo, because 10000 timeout is now multiplied by 1000.
Why are we using seconds instead of milliseconds?

Also docs still says to use 0.
https://nextcloud.github.io/nextcloud-dialogs/interfaces/_toast_.toastoptions.html#timeout...

@nickvergessen @ChristophWurst

skjnldsv:
Fixed by https://github.com/nextcloud/nextcloud-dialogs/pull/210

PR #210, opened 2020-10-08
Export default undo timeout and fix timeout unit

skjnldsv:
Fix timeout unit to use the same as toastify.
We also get closer to the syntax of `setTimeout`, let's not mix units in js :see_no_evil: 
___
Allows us to define fancy setTimeouts

```js
import { showUndo, TOAST_UNDO_TIMEOUT } from '@nextcloud/dialogs'

// DELETION
onDeleteWithUndo() {
	this.deleted = true
	const timeOutDelete = setTimeout(this.onDelete, TOAST_UNDO_TIMEOUT)
	showUndo(t('comments', 'Comment deleted'), () => {
		clearTimeout(timeOutDelete)
		this.deleted = false
	})
},
```

### DIFF
--- a/lib/toast.ts
+++ b/lib/toast.ts
@@ -36,7 +36,7 @@ export const TOAST_UNDO_TIMEOUT = 10000
 
 export interface ToastOptions {
 	/**
-	 * Defines the timeout after which the toast is closed. Set to -1 to have a persistent toast.
+	 * Defines the timeout in milliseconds after which the toast is closed. Set to -1 to have a persistent toast.
 	 */
 	timeout?: number
 
@@ -80,7 +80,7 @@ export interface ToastOptions {
  */
 export function showMessage(data: string|Node, options?: ToastOptions): Toast {
 	options = Object.assign({
-		timeout: 7,
+		timeout: 7000,
 		isHTML: false,
 		type: undefined,
 		// An undefined selector defaults to the body element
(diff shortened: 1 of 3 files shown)


juliushaertl (approved):

lib/toast.ts
```diff
@@ -32,9 +32,12 @@ class ToastType {
 	static readonly UNDO = 'toast-undo';
 }
 
+export const TOAST_UNDO_TIMEOUT = 10000
+export const TOAST_DEFAULT_TIMEOUT = 7000
+
 export interface ToastOptions {
 	/**
-	 * Defines the timeout after which the toast is closed. Set to -1 to have a persistent toast.
+	 * Defines the timeout in milliseconds after which the toast is closed. Set to -1 to have a persistent toast.
```
nickvergessen, line 40: I would recommend a major version bump because of this, as it "breaks" the old usages, well not per se, but a 7ms pop up will be missed, so it is.

lib/toast.ts
```diff
@@ -32,9 +32,12 @@ class ToastType {
 	static readonly UNDO = 'toast-undo';
 }
 
+export const TOAST_UNDO_TIMEOUT = 10000
+export const TOAST_DEFAULT_TIMEOUT = 7000
```
nickvergessen, line 36: While on it add a …?
```js
export const TOAST_PERMANENT = -1
```

skjnldsv: Sure!

nickvergessen (approved):
Sure

PR #222, opened 2020-10-20
Add dark mode close button

PVince81:
Because before this it was black on black and not visible.

Regular:
image

Hover:
image


ChristophWurst (approved):

nickvergessen:
No stable branches here.

styles/toast.scss
```diff
@@ -100,3 +100,15 @@
 		border-left: 3px solid var(--color-success);
 	}
 }
+
+/* dark theme overrides */
+.theme--dark {
```
nickvergessen, line 105: `body.theme--dark`?

PR #236, opened 2020-11-12
Increase toast z-index above modal dialogs

PVince81:
Because our modals are on 1000 and we want to see the toast notifications higher: https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/Modal/Modal.vue#L496

### DIFF
--- a/styles/toast.scss
+++ b/styles/toast.scss
@@ -32,7 +32,7 @@
 	padding-right: 34px;
 	margin-top: 45px;
 	position: fixed;
-	z-index: 9000;
+	z-index: 10100;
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):

skjnldsv:
@PVince81 you forgot to update the changelog https://github.com/nextcloud/nextcloud-dialogs/blob/master/CHANGELOG.md

Only @nc/vue doesn't have it (maybe we should), I think Chrisoph's libraries uses them and Christoph prefers to have it via the Changelog instead of the release list :wink:

PR #340, opened 2021-04-21
Docs about svg bc break

zorn-v:
Closes https://github.com/nextcloud/nextcloud-dialogs/issues/338


ChristophWurst (approved):
Thanks a lot!

PR #423, opened 2021-08-17
fix overlapping menu

szaimen:
Fix https://github.com/nextcloud/server/issues/26518

| Before | After |
|---|---|
| image | image |

Signed-off-by: szaimen <szaimen@e.mail.de>

### DIFF
--- a/styles/toast.scss
+++ b/styles/toast.scss
@@ -32,7 +32,7 @@
 	padding-right: 34px;
 	margin-top: 45px;
 	position: fixed;
-	z-index: 10100;
+	z-index: 2000; // Needs to be below Nextcloud menu entries see https://github.com/nextcloud/server/issues/26518
 	border-radius: var(--border-radius);
 
 	.toast-undo-button,


styles/toast.scss
```diff
@@ -32,7 +32,7 @@
 	padding-right: 34px;
 	margin-top: 45px;
 	position: fixed;
-	z-index: 10100;
+	z-index: 2000; // Needs to be below Nextcloud menu entries see https://github.com/nextcloud/server/issues/26518
```
juliushaertl, line 35: This will cause a regression with modals as the 10100 value was introduced in https://github.com/nextcloud/nextcloud-dialogs/pull/236 to fix that

juliushaertl (changes requested):

styles/toast.scss
szaimen: for which modals would this apply?

juliushaertl: Any modal from the vue library for example: viewer, firstrunwizard, settings dialogs

szaimen: So in this case do you see a way to fix this https://github.com/nextcloud/server/issues/26518 at all?

juliushaertl: > So in this case do you see a way to fix this nextcloud/server#26518 at all?

Can't think of a proper way currently. The issue is that the modal should overlap all content (also the header menus) while the toast should still be overlapped by the menus.

> maybe can we move the toasts a bit to the left? so that it doesn't overlap anymore with the menu?

Also not a solution as the menu has a dynamic width depending on the translations and the same issue also exists for the other popovers like search.

jancborchardt:
Design-wise it looks good! But yeah, Julius’ feedback is valid :)

styles/toast.scss
szaimen: Is there maybe a way to have the z-index for the fialogs by default on 2000 and simply increase it to 10100 when the modal opens and decrease it again when it gets closed?

nickvergessen: > Is there maybe a way to have the z-index for the dialogs by default on 2000 and simply increase it to 10100 when the modal opens and reset it again when it gets closed?

Not sure how that would solve anything. Notifications e.g. is a popup just like in the screenshot, but the toasts triggered from it should overlap the popup. So the library can not solve this. In the case above we should:
1. Make sure the quota error is only shown once
2. The user should use the close button to navigate to the menu.

skjnldsv:
What is the status here?
We have 4.0.0 major on the line, in case this needs to get in as a breaking change :)

szaimen:
> What is the status here? We have 4.0.0 major on the line, in case this needs to get in as a breaking change :)

It can be closed as this is not fixable here.

skjnldsv:
Thanks for the update @szaimen :)

PR #576, opened 2022-05-02
Prevent side effects when importing l10n.js

artonge:
When `@nextcloud/dialogs` is imported, `detectLocale` is called, which try to access `OC`. In some cases, `OC` might not be set yet.

This PR simply defer the `l10n` initialization to when the `n` and `t` functions are actually called.

Not sure if this is the cleanest way :thinking:

### DIFF
--- a/lib/l10n.js
+++ b/lib/l10n.js
@@ -1,12 +1,52 @@
 import { getGettextBuilder } from '@nextcloud/l10n/dist/gettext'
 
-const gtBuilder = getGettextBuilder()
-	.detectLocale()
+let _n, _t = null
 
-process.env.TRANSLATIONS
-	.map(data => gtBuilder.addTranslation(data.locale, data.json))
+// Used to prevent side effects when importing.
+function initL10n() {
+	if (_n !== null) {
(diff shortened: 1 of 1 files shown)


ChristophWurst:
The cleanest way would be to get rid of the global variable dependency and inline that logic

artonge:
> The cleanest way would be to get rid of the global variable dependency and inline that logic

Not sure what you mean. Are you proposing to rebuild `n` and `t` on every call?

ChristophWurst:
The package isn't standalone. It only works when some global variables exist. Therefore the loading orders are so important. If we can manage to get rid of the usage of any global variable then the package will work independently.

Maybe https://github.com/nextcloud/nextcloud-files/pull/111 can act as an example.

artonge:
> The package isn't standalone. It only works when some global variables exist. Therefore the loading orders are so important. If we can manage to get rid of the usage of any global variable then the package will work independently.

Oh! You mean the `OC` access? Yes, sure, I'll see what I can do.

Thanks for the example :)

PR #582, opened 2022-05-12
Fix undo toast rendering

artonge:
When a long text is provided for the undo toast, the text gets mixed with the button. It was fine for the close button, as `padding-right: 34px;` was just right for it.

This PR moves to flex layout to make the toast more responsive.

+ Add `display: flex` to the toast and to the undo button container.
+ Remove the `position: absolute` for the close and undo buttons, so they get placed by the flex algo.
+ Add some `margin-right` to the undo button because it was to close of the text.
+ Change the padding to `6px 12px` as `12px` top and bottom is too much when the close and undo button are not placed absolutely.

:warning: The `toast.scss` is loaded multiple time, so we will have to wait until all `@nextcloud/` packages update to the latest version to see some improvements without `!important`.

Before | After
-- | --
Screenshot from 2022-05-12 14-33-25 | Screenshot from 2022-05-12 14-39-08


GretaD (approved):

GretaD:
this fixes the undo when you delete a comment on files too

PR #622, opened 2022-07-20
Set aria-live default and allow to overwrite it

### LINKED ISSUE #255: Notifications should be read by screen reader
I think toast notifications should be read by screen readers when they appears.
Maybe there's a way to integrate this into the component.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alert_role

The "polite" attribute could be set depending whether it's an error or warnings.

In some cases like in Talk, not sure if messages like "connection to signaling server is taking longer" is worth reading out. Maybe need an extra flag to prevent reading those ?

nickvergessen:
Tested with iOS "voice over" and it read the error message now.

Fix #255

### DIFF
--- a/l10n/messages.pot
+++ b/l10n/messages.pot
@@ -2,6 +2,6 @@ msgid ""
 msgstr ""
 "Content-Type: text/plain; charset=UTF-8\n"
 
-#: lib/toast.ts:193
+#: lib/toast.ts:223
 msgid "Undo"
 msgstr ""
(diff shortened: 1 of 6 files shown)


juliushaertl (approved):

nickvergessen:
Let's wait for the new version of https://github.com/apvarun/toastify-js/pull/106

nickvergessen:
v1.12.0 was released and I pushed the dependency update.

PR #691, opened 2022-10-28
Url encode svg and ship it

skjnldsv:
Found when checking the server's css.
We do not import or copy the assets, which means we're just importing the css without any changes to the url resolving
https://github.com/nextcloud/server/blob/99191167167dd6c98dc3ae9b0eca947526e7939f/core/css/server.scss#L10

By encoding this, we're making sure every release here will works out of the box

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,12 +16,12 @@ npm i -S @nextcloud/dialogs
 
 ```js
 import { showMessage, showInfo, showSuccess, showWarning, showError } from '@nextcloud/dialogs'
-import '@nextcloud/dialogs/styles/toast.scss'
+import '@nextcloud/dialogs/dist/index.css'
 ```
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -16,7 +16,7 @@ npm i -S @nextcloud/dialogs
 
 ```js
 import { showMessage, showInfo, showSuccess, showWarning, showError } from '@nextcloud/dialogs'
-import '@nextcloud/dialogs/styles/toast.scss'
+import '@nextcloud/dialogs/dist/index.css'
```
nickvergessen, line 19: 🙈

Break all the apps!

Bump major version?

skjnldsv: Yep, will be a major

ChristophWurst (approved):
:elephant:

PR #694, opened 2022-10-28
Do not change css class names when building

skjnldsv:
Glad we did some beta lol, it seems module do some weird stuff

| Before | After |
|:---------:|:------:|
|image|image|

Ref https://github.com/nextcloud/server/pull/34868

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -15,8 +15,6 @@ import BabelLoaderExcludeNodeModulesExcept from 'babel-loader-exclude-node-modul
 import { DEFAULT_EXTENSIONS } from '@babel/core'
 const extensions = [...DEFAULT_EXTENSIONS, '.ts', '.tsx']
 
-const packageJson = require('./package.json');
-
 const translations = fs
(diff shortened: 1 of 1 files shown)


skjnldsv:
Bump p lease so we can fix server :grin:

szaimen (approved):

ChristophWurst (approved):
Mind shedding some light on why esm broke this and how your patch fixes this?

:+1: :elephant:

skjnldsv:
> Mind shedding some light on why esm broke this and how your patch fixes this?

esm is not needed, the `\module:true` is what caused it.
Keep in mind this is a followup to the beta changes and the scss compilation recently added from https://github.com/nextcloud/nextcloud-dialogs/pull/691

#691 was part of the answer on how to properly compile scss with rollup.
This pr is the last missing part

ChristophWurst:
Ok but I still can't follow.

skjnldsv:
> Ok but I still can't follow.

This is about the scss->css compilation, no about the js bundles

PR #724, opened 2022-12-14
Update rollup and its dependencies to new major `3.7`

### LINKED ISSUE #696: One word translation madly increase bundle size
If you include `@nextcloud/dialogs` in you project, compiled js size is increased by ~120kb. 
This is insane.
Only `env.TRANSLATION` is about 42kb for one word. Also some size is added for `getText` dances.
I do not need it at all in my project, I even does not use `showUndo` function.

Check complied js in dist please.

susnux:
* This updates rollup to new major 3
  * This supersedes #720
* Fixes the script `check-types` to not emit files
* Fixes #696 by allowing tree shaking as the translations are now build in by replacement instead of process environment

The last item is the main reason why I created this PR, as currently builds of apps using `vite` or `rollup` and depending on this package do not work as then `TRANSLATIONS` are bundled like this `({}).TRANSLATIONS.map`.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -2,14 +2,14 @@
     "compilerOptions": {
       "allowSyntheticDefaultImports": true,
       "declaration": true,
-      "declarationDir": "dist",
       "esModuleInterop": true,
+      "outDir": "./dist",
       "lib": ["ES5", "ES6", "dom"],
       "module": "ESNext",
       "moduleResolution": "node",
       "rootDir": "lib/",
       "sourceMap": true,
-      "target": "ES5"
+      "target": "ES2015"
     },
     "include": ["lib/*"],
     "exclude": [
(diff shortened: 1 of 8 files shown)


skjnldsv (approved):

tsconfig.json
```diff
@@ -2,14 +2,14 @@
     "compilerOptions": {
       "allowSyntheticDefaultImports": true,
       "declaration": true,
-      "declarationDir": "dist",
+      "declarationDir": "./types",
```
skjnldsv, line 5: We never did this anywhere in other libs.
Is there any reason for it?

skjnldsv: Nice! Make sense :)
I actually noticed that not long ago in another location, don't recall where :thinking:

PR #760, opened 2023-02-14
feat: Make dependencies external

susnux:
`dependencies` should be external and only built into the bundle if the target is `browser`.

Also as the `type` is set to `module` config files are not longer required to be named `mjs` instead `js` is sufficient.


nickvergessen:
Removing myself as I have no clue of that code level

juliushaertl (approved):

PR #761, opened 2023-02-15
Fix package resolving in webpack projects

susnux:
Currently the *4.0.0* release can not be used with webpack projects, because they will throw:
`ReferenceError: exports is not defined`.

Webpack tries to resolve the import of `@nextcloud/dialogs` to `node_modules/@nextcloud/dialogs/dist/index.js`.
This is a commonjs file, but webpack expects an esm file, so `exports` is not defined.

I guess this happens because `dialogs` has its type set to `module´ so every `.js` file within will be handled as esm.

So to fix this we need to explicitly name commonjs files `.cjs`.

---
I noticed this issue when trying to use `@nextcloud/vue 7.6.0` which depends on dialogs version `4.0.0`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,13 +3,13 @@
   "version": "4.0.0",
   "description": "",
   "types": "dist/index.d.ts",
-  "main": "dist/index.js",
-  "module": "dist/index.es.js",
+  "main": "dist/index.cjs",
+  "module": "dist/index.mjs",
   "type": "module",
   "exports": {
     ".": {
-      "import": "./dist/index.es.js",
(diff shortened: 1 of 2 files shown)


mejo- (approved):
I can verify that this fixes using `@nextcloud/vue` 7.6.0.

ChristophWurst (approved):

juliushaertl:
Thanks @susnux I'll spin a new release

juliushaertl:
Building at https://github.com/nextcloud/nextcloud-dialogs/actions/runs/4193918096

PR #803, opened 2023-04-28
Add an ability to set and reset Toastify options globally

### LINKED ISSUE #380: Global toastify options
Make it possible to set toastify options globally so that one doens't need to pass in the same options everywhere.

One use case would be to pass in an offset for apps that have a top bar, see https://github.com/nextcloud/spreed/pull/5794
or even to switch to bottom aligned notifications for some layouts.

Antreesy:
### ☑️ Resolves

* Add `setGlobalToastOptions(options: ToastOptions)` function
* Add `resetGlobalToastOptions()` function
* Update README.md
* Fix typos, remove redundant comments
* Close #380 

___

### 🚧 Tasks

- [ ] Code review
- [x] ⛑️ Tested in sandbox and Talk
- [ ] 📘 API or User documentation - should be updated?

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,12 @@ There are several options that can be passed in as a second parameter, like the 
 showError('This is an error shown without a timeout', { timeout: -1 })
 ```
 
+You could set option for the whole project globally:
+
+```
+setGlobalToastOptions({ selector: 'vue-app',  timeout: -1 })
+```
+
 A full list of available options can be found in the [documentation](https://nextcloud.github.io/nextcloud-dialogs/).
 
 ## Releasing a new version
--- a/l10n/messages.pot
+++ b/l10n/messages.pot
@@ -2,6 +2,6 @@ msgid ""
 msgstr ""
 "Content-Type: text/plain; charset=UTF-8\n"
 
-#: lib/toast.ts:223
+#: lib/toast.ts:243
 msgid "Undo"
 msgstr ""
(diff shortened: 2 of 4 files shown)


ChristophWurst:
This makes it still local to the app

Antreesy:
> This makes it still local to the app

That shouldn't be the problem, if we're going to define a container separately for each application.

For example, in Files app container is `<main id="content" class="app-files">`,
 when in Talk it is `<div  id="content-vue" class="content app-talk">`.
 
So if we manage to set the global selector for the whole NC instance, it's still needs to be manually changed for Talk, which makes not much sense in that case

ShGKme:
> This makes it still local to the app

The question is, do we want to make it globally across the apps? 

1. On one hand, it may solve a problem then toasts from one app don't work because of the layout of another app.
2. On the other hand, in this case, one app changes the behavior of another app in a possibly unexpected way.

I'm not sure if p.1 is an actual problem. 

What about to start with a module-wide global and then if there will be a case, to add a fully global option?

nickvergessen:
> On one hand, it may solve a problem then toasts from one app don't work because of the layout of another app.

That is actually one of the problems we try to solve.
Currently when the Talk fullscreen option is used, the integrations of deck, tasks, etc. all miss their toasts as they are still displayed on the body, instead of the fullscreened container.

ShGKme:
Then can we store the value for the default selector in the `window`, similar to `@nextcloud/l10n`?

https://github.com/nextcloud/nextcloud-l10n/blob/master/lib/registry.ts#L69

PR #843, opened 2023-06-21
fix: Make sure dependencies are external

susnux:
Rollup does check strings for full match, so transforming our dependencies list to RegExp will work even with imports like `import 'core-js/foo/bar`.

Before: `1'824 kB` (yes 1.8 MiB)
After: `237 kB`

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -23,7 +23,10 @@ const packageJson = JSON.parse(
 	)
 )
 
-const externals = [...Object.keys(packageJson?.dependencies || {}), ...Object.keys(packageJson?.peerDependencies || {})]
+const externals = [
+	...Object.keys(packageJson?.dependencies || {}),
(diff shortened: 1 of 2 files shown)


Pytal (approved):

rollup.config.js
```diff
@@ -23,7 +23,7 @@ const packageJson = JSON.parse(
 	)
 )
 
-const externals = [...Object.keys(packageJson?.dependencies || {}), ...Object.keys(packageJson?.peerDependencies || {})]
+const externals = [...Object.keys(packageJson?.dependencies || {}), ...Object.keys(packageJson?.peerDependencies || {})].map(v => new RegExp(`^${v}`))
```
Pytal, line 26: I believe `v` is the package name so should we rename it to something like `package`, as v makes it seem like the version?

PR #863, opened 2023-07-12
chore: Do not bundle `core-js` polyfills already included in server core

susnux:
We do not need to include core-js polyfills, they are already bundled with the server core.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -4,8 +4,7 @@ export default {
         [
             "@babel/env",
             {
-                useBuiltIns: "usage",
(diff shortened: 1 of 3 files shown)


skjnldsv (approved):

skjnldsv:
Theoretically we shouldn't even use that config but `@nextcloud/babel-config`

nickvergessen:
Removing myself as I have no JS package knowledge

ChristophWurst (approved):

PR #878, opened 2023-08-10
Implement Vue based Dialogs especially the FilePicker

### LINKED ISSUE #1: Ship own filepicker code
Right now we just delegate, but ideally this package should be independent

### LINKED ISSUE #301: Vue-Rewrite?
I recently thought about implementing #300 just like the filePicker. But probably, and in combination with #1 it would be better to first rewrite the own library, before adding new stuff?
Maybe some nice task for the vue-expert @skjnldsv ? 🙈

susnux:
* Resolves #1 
* Resolves #301

### Done :heavy_check_mark: 
* [x] Write reusable Vue Dialog component
* [x] Implement FilePicker as a Vue Dialog
* [x] Implement suggested improvements
  * [x] Remember last path (https://github.com/nextcloud/server/issues/39186)
  * [x] A similar layout to the settings modal is used to mimic the files app layout, with a navigation on the left containing
    * [x] All files
    * [x] Recent
    * [x] Favorites
  * [x] If "All files" is chosen, the breadcrumbs are shown with the file list
  * [x]If "Recent" is chosen, a heading "Recent items" is shown and the recent files list
  * [x] If "Favorites" is chosen, a heading "Favorites" and the list of favorites
  * [x] The new breadcrumbs component is used
    * :warning: There are errors in the console, but they seem to come from the vue component itself.
  * [x] Remove the lines in between each item
  * [x] https://github.com/nextcloud/server/issues/39187 : the first item of the navigation can be a search bar which allows you to search for a file (across all your files)
* [x] Loading placeholders similar to the text app
* [x] Write legacy wrapper so the new FilePicker can be used with the same API, we just need to switch the function call in  server core.

### Todo :construction: 
(in a follow up)
* [ ] Move the Dialog and DialogButton to `@nextcloud/vue`
* [ ] Implement Grid view
* [ ] Load view order from files app API (https://github.com/nextcloud/server/issues/39185)
* [ ] Proper tests
* [ ] *(Not related but still a todo :wink: )* Enforce eslint rules

### Screenshots

image

https://github.com/nextcloud-libraries/nextcloud-dialogs/assets/1855448/3f5ec8fe-ec11-4308-824f-5dca90c72752

And on smaller screens:
image


susnux:
cc @nimishavijay

AndyScherzinger (approved):

skjnldsv:
Follow-up: checkboxes should be hidden when picking a destination?

skjnldsv (approved):

skjnldsv:
Insane work btw!!

AndyScherzinger:
Awesome work @susnux - left open is then the server-PR to make use of it via a lib bump I guess?

raimund-schluessler:
>    * [x]  The new breadcrumbs component is used
>
>      * ⚠️ There are errors in the console, but they seem to come from the vue component itself.

Nice to see the component is finally being used somewhere. Are the error something I can help with?

skjnldsv:
> Nice to see the component is finally being used somewhere

Also in the new Files app :rocket:

susnux:
@raimund-schluessler 
> Are the error something I can help with?

The mounted hook fails because `getWidth` throws an error, my current guess is that the function uses `.elm` on vnodes which is not always defined. But I will open an issue on the vue repository with some tracing :)

susnux:
@AndyScherzinger 
> Awesome work @susnux - left open is then the server-PR to make use of it via a lib bump I guess?

Thank you, yes we need a new (4.2.0-beta.1 ?) release. The PR is already prepared (~6 lines of code to change) so that is used everywhere :)

AndyScherzinger:
Yes, can you create a tag? @skjnldsv, me or any team lead can then create the release to trigger the npm publishing part.

juliushaertl:
Wow. So nice to see this finally happend. Great work @susnux 💙

raimund-schluessler:
Is there some documentation how to use the new filepicker? I try with an app that used the old filepicker like this:
```js
const file = await getFilePickerBuilder('Pick a file').build().pick()
```
but that still leads to the old picker.

And using the new picker directly via `import { FilePickerVue } from '@nextcloud/dialogs'` and e.g. showing it with `v-if` leads to an error regarding buttons:
```
TypeError: a.buttons is undefined
    r FilePicker-d6e48816.mjs:166
    VueJS 3
    i FilePicker-d6e48816.mjs:166
    VueJS 4
    xt FilePicker-d6e48816.mjs:185
    VueJS 32
    ru Attachments.vue:1
    VueJS 7
vue.runtime.esm.js:3049:8
```

I obviously do something wrong here, so the documentation would be helpful :slightly_smiling_face:

AndyScherzinger:
@susnux knows best and already integrated it in the files app.

AndyScherzinger:
See https://github.com/nextcloud/server/pull/39792 for an example @raimund-schluessler

susnux:
> Is there some documentation how to use the new filepicker?

Currently not, vue files broke Typedoc. I will have to figure out how to document vue files with typedoc. Planned this for the next dialogs release.

> I try with an app that used the old filepicker like this:
> 
> ```js
> const file = await getFilePickerBuilder('Pick a file').build().pick()
> ```
> 
> but that still leads to the old picker.

Yes the FilePicker builder still relays on the OC.dialogs variable so only shows the new FilePicker for 27.1.0-beta currently.
This is also planned to migrate to always use the provided filepicker instead of using the global variable.

> And using the new picker directly via `import { FilePickerVue } from '@nextcloud/dialogs'` and e.g. showing it with `v-if` leads to an error regarding buttons:
> ...
> I obviously do something wrong here, so the documentation would be helpful 🙂

You are missing the buttons property which currently (!) looks like this:
```ts
type IFilePickerButtons = Array<IFilePickerButton>

interface IFilePickerButton {
    // The button text
    label: string
    // Optional button icon (as a Vue component), recommend to use the AsyncComponent factory
    icon?: Component | AsyncComponen
    // Same as NcButton
    type: string
    // callback on click
    callback: (nodes: Node[]) => void
}
```

raimund-schluessler:
> Yes the FilePicker builder still relays on the OC.dialogs variable so only shows the new FilePicker for 27.1.0-beta currently. This is also planned to migrate to always use the provided filepicker instead of using the global variable.

That is weird, I tried it with latest master of server, but still the old picker.

> You are missing the buttons property which currently (!) looks like this:
> 
> ```ts
> type IFilePickerButtons = Array<IFilePickerButton>
> 
> interface IFilePickerButton {
>     // The button text
>     label: string
>     // Optional button icon (as a Vue component), recommend to use the AsyncComponent factory
>     icon?: Component | AsyncComponen
>     // Same as NcButton
>     type: string
>     // callback on click
>     callback: (nodes: Node[]) => void
> }
> ```

Thanks for the hint. I am not used to the composition api yet, but I will try to figure out how to provide the buttons property.

susnux:
> That is weird, I tried it with latest master of server, but still the old picker.

Yes currently only in 27.1.0 beta, master PR is still pending :/

> Thanks for the hint. I am not used to the composition api yet, but I will try to figure out how to provide the buttons property.

Just provide a the Filepicker the buttons, like
```vue
<template>
<FilePicker :buttons="buttons" ... />
</template>
<script>
// ...
export default {
 data() {
    return {
        buttons: [{ label: 'Pick', callback: (nodes) => console.log('picked nodes:', nodes) }]
    }
  }
// ...
}
</script>
```

raimund-schluessler:
> @raimund-schluessler
> 
> > Are the error something I can help with?
> 
> The mounted hook fails because `getWidth` throws an error, my current guess is that the function uses `.elm` on vnodes which is not always defined. But I will open an issue on the vue repository with some tracing :)

@susnux See https://github.com/nextcloud-libraries/nextcloud-vue/pull/4416

PR #880, opened 2023-08-10
fix(FilePicker): Adjust color of placeholder skeletons

susnux:
Adjust color as suggested by @jancborchardt (I can somehow not request a review from you using GH, but feedback is very welcome :wink: ).

I changed the background to
```css
background: linear-gradient(to right,var(--color-background-darker),var(--color-text-maxcontrast),var(--color-background-darker));
background-size: 600px 100%;
border-radius: var(--border-radius);
animation: gradient 12s ease infinite;
```

### Screenshots

before | after
---|---
image | image

### DIFF
--- a/lib/components/FilePicker/LoadingTableRow.vue
+++ b/lib/components/FilePicker/LoadingTableRow.vue
@@ -40,7 +40,7 @@
 		display: inline-block;
 		height: 24px;
 
-		background: linear-gradient(to right,var(--color-background-darker),var(--color-main-text),var(--color-background-darker));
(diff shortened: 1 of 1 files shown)


AndyScherzinger (approved):

jancborchardt (approved):
Looks nice! Is there already a component for this btw or some specific variables? If not it would be good to aligncc @skjnldsv @raimund-schluessler

susnux:
@jancborchardt I do not think there is a component for this, not sure if it makes sense as it depends on your real content. Maybe just provide the color / background animation of the skeletons as a variable. But maybe someone has an idea how a flexible component can be created for this?

PR #884, opened 2023-08-11
feat(vue3): migrate to vue3

raimund-schluessler:


raimund-schluessler:
@susnux I worked a bit on the migration, but I could use help with the Typescript stuff.

lib/components/FilePicker/FilePickerBreadcrumbs.vue
```diff
@@ -25,7 +25,7 @@
 					<IconPlus :size="20" />
 				</template>
 				<NcActionInput ref="nameInput"
-					:value.sync="newNodeName"
+					v-model:value="newNodeName"
```
susnux, line 28: We probably should rename this on `@nextcloud/vue` to `modelValue` and just use `v-model`, no?

raimund-schluessler: It is already. It's a typo here.

raimund-schluessler: Fixed.

Antreesy:
Hi @susnux!
i rebased the branch and tried to resolve conflicts from recent app changes. TSC and tests are failing, so I might have not done it correctly 🙈.
Could you double-check, when you have time?

ShGKme:
@raimund-schluessler I'll take over the PR to prepare the release soon, if it's ok

raimund-schluessler:
Great, please go ahead!

lib/utils/dialogs.ts
```diff
@@ -13,23 +13,20 @@ import Vue, { toRaw } from 'vue'
...
-				on: {
-					close: (...rest: unknown[]) => {
-						onClose(...rest.map(v => toRaw(v)))
-						vue.$destroy()
-					},
+				onClose: (...rest: unknown[]) => {
+					onClose(...rest.map(v => toRaw(v)))
+					vue.unmount()
 				},
 			}),
```
st3iny, line 31: AFAIK, this won't work in Vue 3. We can only pass `propsData` as the second argument. We need to find another workaround to wrap the onClose handler.

We also need to mount the component explicitly here.

## My proposed solution:

```js
	const vue = createApp(dialog, props)
	vue.mount(el)
```

st3iny (changes requested):
Seems to work otherwise.

lib/utils/dialogs.ts
ShGKme: > We can only pass `propsData` as the second argument

This is true for `createApp().mount()`, not for `h`

Problem here is that props should be destructured.

ShGKme: > We need to find another workaround to listen to the close event or simply remove the `onClose()` handler.

In Vue 3 event handlers always passed as `onEven` param.

`<comp @event="handler" />` is `h(comp, { onEvent: handler })`

st3iny: TIL. I wasn't aware of the advanced features of `h`.

susnux: for dialogs library this `createApp` is fine, but for nextcloud-vue it can cause unwanted behavior:
- it does not allow globally registered components in the dialog
- it does not allow to access registered pinia

Use case would be e.g. a settings store for dialog you spawn with API.

PR #921, opened 2023-08-23
fix(FilePicker): The table height should be max. 100%

susnux:
If the height is set to 100% the rows are resized and look weird.

### Screenshots
before | after
---|---
image|image


szaimen (approved):
Lgtm from the screemshot

PR #923, opened 2023-08-23
[stable4] Add unittests for usables

susnux:
Manual backport of #917


susnux:
I let this open, as this might have the same problems on server as https://github.com/nextcloud/server/pull/39955

susnux:
> I let this open, as this might have the same problems on server as https://github.com/nextcloud/server/pull/39955

Should be ok now with beta19 of files library.

susnux:
Well no. We use nextcloud-vue 7 which only provides CJS entry points, those need to be converted to ESM but somehow the `require('vue-material-design-icons/...')` are not handled by the vue plugin but directly, so that there are invalid syntax errors.

susnux:
lets not do

Issue #920, opened 2023-08-22
Design review


susnux:
Copy-Pasted notes from design review to be triaged and implemented:
* [x] Add button should look like in files app (#928)
* [x] Remove checkboxes for single select (https://github.com/nextcloud-libraries/nextcloud-dialogs/pull/929)
* [x] If there are no files in a folder, show EmptyContent component (https://github.com/nextcloud-libraries/nextcloud-dialogs/pull/930)
* [x] Hover on checkboxes should be a circle (Vue component issue) (https://github.com/nextcloud-libraries/nextcloud-vue/pull/4453)
* [x] Add icons to copy and move by default (#938 )
* [x] Move --> Move to [foldername] (#938 )
* [x] Change folder icon size to 20px (todo with F2V) (https://github.com/nextcloud-libraries/nextcloud-dialogs/pull/962)
* [x] heading for the entire dialog should be center aligned with the dialog --> part of dialogs rework (https://github.com/nextcloud-libraries/nextcloud-dialogs/pull/965)
  * [x] Have a heading or title prop, centered in the dialog 
* [x] Hover feedback for "home" should be circle --> Vue component, so forwarded in https://github.com/nextcloud-libraries/nextcloud-vue/issues/4519
* [x] checkboxes should not be shown above the header when you scroll (z-index of checkboxes too high) (https://github.com/nextcloud-libraries/nextcloud-dialogs/pull/997)
* [x] Sorting --> last sorting from user in files app
  * [x] Have a default sorting by name for the time being (#949)

Descoped
* [ ] Exchange positions of text field and dropdown in "medium" screen size 
* [ ] Position of search box on the right? +  functionality of search --> to discuss
  * If the search does only filter it should be placed on the right of the bread crumbs, expandable like the talk search
  * If it really searches it can stay where it is

susnux:
> Add icons to copy and move by default

Can someone recommend icons for this purpose?
I am not happy with the ones I found on MDI :sweat_smile:  @szaimen @nimishavijay

skjnldsv:
> Can someone recommend icons for this purpose?

Move is [`folder-move`](https://pictogrammers.com/library/mdi/icon/folder-move/)
Copy is [`content-copy`] (https://pictogrammers.com/library/mdi/icon/content-copy/)

skjnldsv:
> But they're not available in our `@mdi/svg` or `vue-material-design-icons` libs it seems 🤔

[...]

PR #929, opened 2023-08-23
fix(FilePicker): Do not show checkboxes on single select mode

susnux:
Selected file is only displayed by background color and checkboxes are hidden if `multiselect` is `false`.

Screen Shot 2023-08-23 at 19 02 38

Part of #920


skjnldsv:
Fix #940

PR #930, opened 2023-08-23
Add empty content when there are no files

susnux:
* Part of #920 

If there are no files in the current view or no matching filter, show the empty content component instead of the file list.

image


skjnldsv (approved):

susnux:
still no bot :(

PR #931, opened 2023-08-23
feat(FilePicker): Always show file extension as a trailing text part

susnux:
Same as the file list of the files app, this now shows the file extension as a trailing text part.

image

### DIFF
--- a/lib/components/FilePicker/FileListRow.vue
+++ b/lib/components/FilePicker/FileListRow.vue
@@ -12,6 +12,7 @@
 			<div class="file-picker__name-container">
 				<div class="file-picker__file-icon" :style="{ backgroundImage }" />
 				<div class="file-picker__file-name" :title="displayName" v-text="displayName" />
+				<div class="file-picker__file-extension" v-text="fileExtension" />
 			</div>
 		</td>
 		<td class="row-size">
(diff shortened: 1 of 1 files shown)


skjnldsv (approved):

susnux:
@skjnldsv Do you think the displayname handling is ok, or do we need special extension handling even if the display name is set?

lib/components/FilePicker/FileListRow.vue
```diff
@@ -47,9 +48,14 @@ const emit = defineEmits<{
 }>()
 
 /**
- * The displayname of the current node
+ * The displayname of the current node (excluding file extension)
  */
-const displayName = computed(() => props.node.attributes?.displayname || props.node.basename)
+const displayName = computed(() => props.node.attributes?.displayname || props.node.basename.slice(0, props.node.extension ? -props.node.extension.length : undefined))
```
skjnldsv, line 53:
```suggestion
const displayName = computed(() => props.node.attributes?.displayName || props.node.basename.slice(0, props.node.extension ? -props.node.extension.length : undefined))
```

https://github.com/nextcloud/server/blob/24af920992ba668a90bb71d5a2bce84109a6587e/apps/files/src/components/FileEntry.vue#L323-L336

skjnldsv, line 53:
```suggestion
const displayName = computed(() => props.node.attributes?.displayName || props.node.basename.slice(0, props.node.extension ? -props.node.extension.length : undefined))
```

https://github.com/nextcloud/server/blob/24af920992ba668a90bb71d5a2bce84109a6587e/apps/files/src/components/FileEntry.vue#L323-L336

skjnldsv (dismissed):

lib/components/FilePicker/FileListRow.vue
susnux: Thank you!

skjnldsv (approved):

PR #934, opened 2023-08-23
fix(FilePicker): Fix selecting rows using the keyboard

susnux:
Selecting by keyboard was broken due to wrong event and key name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,19 @@ const paths = await filepicker.pick()
 </script>
 ```
 
-## Releasing a new version
+## Development
+### Testing
+For testing all components provide `data-testid` attributes as selectors, so the tests are independent from code or styling changes.
+
+### Test selectors
+`data-testid` | Intended purpose
+---|---
+`select-all-checkbox` | The select all checkbox of the file list
+`file-list-row`       | A row in the file list (`tr`), can be identified by `data-filename`
+`row-checkbox`        | Checkbox for selecting a row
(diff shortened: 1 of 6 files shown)


susnux:
Ref: https://github.com/perry-mitchell/webdav-client/issues/355

skjnldsv (approved):

susnux:
Not sure what to do here. Merge without tests? Is there some way to ignore the package exports? Last resort: Fork the project as like with vue-select?

skjnldsv:
> Not sure what to do here. Merge without tests? Is there some way to ignore the package exports? Last resort: Fork the project as like with vue-select?

Import with relative path.

susnux:
> Import with relative path.

This might work for apps or server, but not for libraries. You can not know that the package is installed at that location.
So for libraries this only works if we bundle the dependency inside the library.

As all of those are possible:
```
server
  |- node_modules
      |- @nextcloud/dialogs
      |- webdav
```
or
```
server
  |- node_modules
      |- @nextcloud/dialogs
          |- node_modules
              |- webdav
```
or
```
server
  |- node_modules
      |- @nextcloud/dialogs
          |- node_modules
              |- @nextcloud/files
                  |- node_modules
                      |- webdav
```

skjnldsv:
Indeed, I only tried with Apps

susnux:
See #949 for a solution until it is fixed, after that one is merged, this can be rebased and should work :)

PR #953, opened 2023-08-25
fix: dependencies and force bundle icons

skjnldsv:
## Why not include `@nextcloud/vue`

All node_modules are already excluded, this config is therefore unecessary

## Why adding those packages to peerDependencies

They are excluded from the output, but required/imported by this library
If we do not specify the required versions, npm will not know what to install and depend to

## Why bundling icons

Because they are static components and it's more stable this way.
It's either this or we add them into the peerDependencies, but considering the dependency mess this can create, it's simpler and steadier to just bundle them imho. The change is not that big anyway (7 KB, uncompressed)

### Before
```sh
dist/index.cjs                        0.95 kB │ gzip: 0.37 kB │ map:  0.09 kB
dist/chunks/FilePicker-bc9a2f03.cjs  20.93 kB │ gzip: 6.47 kB │ map: 51.38 kB
dist/chunks/index-2cfdbca8.cjs       85.41 kB │ gzip: 7.89 kB │ map: 57.93 kB
```

### After
```sh
dist/index.cjs                        0.95 kB │ gzip: 0.37 kB │ map:  0.09 kB
dist/chunks/FilePicker-104c4213.cjs  27.88 kB │ gzip: 7.54 kB │ map: 62.46 kB
dist/chunks/index-0065eb51.cjs       85.77 kB │ gzip: 8.10 kB │ map: 58.64 kB
```

## Why not moving webdav to devDependencies

webdav is outputed to the dist folder. I am still not sure it belongs there.
Considering this is just typings, I do wonder if we really need to expose it 
```js
grep -rn webdav dist 
dist/usables/dav.d.ts:24:import type { WebDAVClient } from 'webdav';
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,12 @@
     "dist"
   ],
   "peerDependencies": {
-    "vue": "^2.7.14"
+    "@nextcloud/files": "^3.0.0-beta.19",
+    "@nextcloud/l10n": "^2.2.0",
+    "@nextcloud/typings": "^1.7.0",
+    "@nextcloud/vue": "^8.0.0-beta.3",
+    "vue": "^2.7.14",
+    "webdav": "^5.2.3"
   },
   "dependencies": {
     "@mdi/svg": "^7.2.96",
@@ -61,18 +66,17 @@
     "@nextcloud/l10n": "^2.2.0",
     "@nextcloud/typings": "^1.7.0",
     "@nextcloud/vue": "^8.0.0-beta.3",
-    "@types/toastify-js": "^1.12.0",
     "@vueuse/core": "^10.3.0",
     "toastify-js": "^1.12.0",
     "vue-frag": "^1.4.3",
-    "vue-material-design-icons": "^5.2.0",
-    "webdav": "^5.2.3"
+    "vue-material-design-icons": "^5.2.0"
   },
   "devDependencies": {
     "@nextcloud/browserslist-config": "^3.0.0",
     "@nextcloud/eslint-config": "^8.3.0-beta.2",
     "@nextcloud/vite-config": "^1.0.0-beta.18",
     "@types/gettext-parser": "^4.0.2",
+    "@types/toastify-js": "^1.12.0",
     "@vitest/coverage-istanbul": "^0.34.2",
     "@vue/test-utils": "^1.3.6",
     "@vue/tsconfig": "^0.4.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,18 +15,17 @@
         "@nextcloud/l10n": "^2.2.0",
         "@nextcloud/typings": "^1.7.0",
         "@nextcloud/vue": "^8.0.0-beta.3",
-        "@types/toastify-js": "^1.12.0",
         "@vueuse/core": "^10.3.0",
         "toastify-js": "^1.12.0",
         "vue-frag": "^1.4.3",
-        "vue-material-design-icons": "^5.2.0",
-        "webdav": "^5.2.3"
+        "vue-material-design-icons": "^5.2.0"
       },
       "devDependencies": {
         "@nextcloud/browserslist-config": "^3.0.0",
         "@nextcloud/eslint-config": "^8.3.0-beta.2",
         "@nextcloud/vite-config": "^1.0.0-beta.18",
         "@types/gettext-parser": "^4.0.2",
+        "@types/toastify-js": "^1.12.0",
         "@vitest/coverage-istanbul": "^0.34.2",
         "@vue/test-utils": "^1.3.6",
         "@vue/tsconfig": "^0.4.0",
@@ -47,7 +46,12 @@
         "npm": "^9.0.0"
       },
       "peerDependencies": {
-        "vue": "^2.7.14"
+        "@nextcloud/files": "^3.0.0-beta.19",
+        "@nextcloud/l10n": "^2.2.0",
+        "@nextcloud/typings": "^1.7.0",
+        "@nextcloud/vue": "^8.0.0-beta.3",
+        "vue": "^2.7.14",
+        "webdav": "^5.2.3"
       }
     },
     "node_modules/@aashutoshrathi/word-wrap": {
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -53,26 +53,30 @@
...
+    "vue": "^2.7.14",
+    "webdav": "^5.2.3"
   },
   "dependencies": {
     "@mdi/svg": "^7.2.96",
     "@nextcloud/files": "^3.0.0-beta.19",
     "@nextcloud/l10n": "^2.2.0",
     "@nextcloud/typings": "^1.7.0",
     "@nextcloud/vue": "^8.0.0-beta.3",
-    "@types/toastify-js": "^1.12.0",
```
susnux, line 64: This breaks Typescript users as it is included in output / dist

skjnldsv: I could not find it in the output, are you sure?

susnux:
> ## Why not include `@nextcloud/vue`
> All node_modules are already excluded, this config is therefore unecessary

Should be ok :)

> ## Why adding those packages to peerDependencies
> 
> They are excluded from the output, but required/imported by this library If we do not specify the required versions, npm will not know what to install and depend to

of cause thats how dependencies work, if you bundle the dependencies they are just development dependencies are you not require to install those on runtime.

`dependencies` are installed by npm and npm will resolve any conflicts, e.g. puts the dependency inside the node_modules if the parent project depends on a different version.

`peerDependencies` are also (npm7) installed by default, but conflicts are not resolved.
Meaning if you have peerDependency vue2 but the parent project uses vue3, vue 2 will not be installed for your package.

So `dependencies` are runtime dependencies, theoretically each package will install their own version of it into its `node_modules`  (npm is "smart" so use parent dependencies if possible). But you always can guranteed import that version of the dependency.

`peerDependencies` are only required if you extend behavior on the parent or the package is a singleton, like Vue2 or React.

TL;DR: What issue do you try to solve? As that are all only dependencies and not required to be installed by using apps.

> ## Why bundling icons
> 
> Because they are static components and it's more stable this way. It's either this or we add them into the peerDependencies, but considering the dependency mess this can create, it's simpler and steadier to just bundle them imho. The change is not that big anyway (7 KB, uncompressed)

Also here, no need for `peerDependencies`. But I agress with `vue-material-design-icons` as those will require users of the library to have a vue compiler in their build chain (as they are uncompiled SFCs).

> ## Why not moving webdav to devDependencies
> 
> webdav is outputed to the dist folder. I am still not sure it belongs there. Considering this is just typings, I do wonder if we really need to expose it

Exactly we use it in production code (so not only build or test code), thats why this dependency is listed in `dependencies`.

package.json
susnux: `index.d.ts`:
```ts
import { default as Toastify_2 } from 'toastify-js';
```

That is implicitly `@types/toastify-js`.

skjnldsv: Fair enough!

skjnldsv:
> TL;DR: What issue do you try to solve? As that are all only dependencies and not required to be installed by using apps.

Example:
1. have this lib require latest @nextcloud/files
2. have server require `@nextcloud/files@beta.16`
3. see it break on server without any warning or error because the dependency is good enough for npm

> Peer dependencies should be used when a package expects a specific version of another package to be present

Right now this is exactly what is happening
This package is NOT compatible with other versions of our external libraries.
If you install `@nextcloud/vue@^7`, this will not throw. But it will also not work.
We have to specify which versions of `@nextcloud/files`, `@nextcloud/vue`... etc we are truly compatible with.

Look at that:
```js
~/git/server@stable27$ npm ls @nextcloud/vue                                                                                   git:(stable27*)
nextcloud@1.0.0 /home/admin/git/server
├─┬ @nextcloud/calendar-availability-vue@0.6.0-alpha1
│ └── @nextcloud/vue@7.12.2 deduped
├─┬ @nextcloud/dialogs@5.0.0-beta.2
│ └── @nextcloud/vue@8.0.0-beta.3
├─┬ @nextcloud/password-confirmation@4.0.4
│ └── @nextcloud/vue@7.12.2 deduped
├─┬ @nextcloud/vue-dashboard@2.0.1
│ └── @nextcloud/vue@3.10.2
├── @nextcloud/vue@7.12.2
└─┬ nextcloud-vue-collections@0.10.0
  └── @nextcloud/vue@3.10.2
```

susnux:
> see it break on server without any warning or error because the dependency is good enough for npm

But thats because of the insufficient version dependency, omitting the `^` on server will force npm to install two versions. (while in beta).

> Look at that, I installed this library on server stable 27, no errors! 🙈

This is also expected, dependencies should not interact with the parent project. So in theory you could use current master of dialogs on server stable27. The problem is we have insufficient css scoping, so the styles conflict and result in visual issues.
(You could prevent this by using css modules instead of css scopes, see https://vuejs.org/api/sfc-css-features.html#css-modules )

Of cause in our case this is not really expected.

But general putting the dependencies into `peerDependencies`, does not feel right. As this requires the user to stick to that specific version of a different dependency.
**Also that would require us to release an new version every time one of our dependencies has a new release**

So for me I am fine with removing `@nextcloud/vue` as it is already included, and bundling `vue-material-design-icons`. But the peerDependencies seem semantically wrong and probably will break using applications later. Instead server should pin beta versions if not compatible.

skjnldsv:
I don't agree with you on this :)
Being able to install a library without any warning that fails the build is an issue that needs resolving.
You can decided to use peerDependencies following major blocking.
This is what most libs are doing

We're as dependent on `vue` `2.x` here than we are on `@nextcloud/vue` `8.x`
And this is a hard requirement

skjnldsv:
> Also that would require us to release an new version every time one of our dependencies has a new release

"every time one of our dependencies has a new MAJOR release" would be the better outcome in my opinion.

susnux:
> Why do you think having vue as a peerDep is more acceptable than another library then?

Because Vue2 is a singleton here, the vue state is shared and will result in unexpected behavior if you have multiple vue versions / instances in the same app.

> example:
>
>    have this lib require latest @nextcloud/files
>    have server require @nextcloud/files@beta.16
>    see it break on server without any warning or error because the dependency is good enough for npm

But this does not change anything for that issue:
Try to install this branch instead. For example stable27 uses beta8, if you install this branch:
>     "@nextcloud/dialogs": "nextcloud-libraries/nextcloud-dialogs#feat/peers"

It also uses beta19 which is breaking:
```
nextcloud@1.0.0
├─┬ @nextcloud/dialogs@5.0.0-beta.2 (git+ssh://git@github.com/nextcloud-libraries/nextcloud-dialogs.git#5dc7d8e2311c0dabbfaf465b0e409d1ed559188f)
│ └── @nextcloud/files@3.0.0-beta.19 deduped
└── @nextcloud/files@3.0.0-beta.19
```

---

We can go this way if you really want to, I do not like it as for me that are all dependencies and not dependencies a library user should also always have installed. But I do not want to be a pighead, so my opinion about both dependency types is not a blocker.

**But** if you want to solve the issue here that

skjnldsv:
> Because Vue2 is a singleton here, the vue state is shared and will result in unexpected behavior if you have multiple vue versions / instances in the same app.

And we also have unexpected behavior with multiple versions of `@nextcloud/vue` :thinking:

PR #971, opened 2023-08-28
Fix styling issues and make the FilePicker height fixed

### LINKED ISSUE #900: Question: Should the file picker have a fixed size?
Currently the file picker is not fixed in height, it is just clamped to a minimum of 4 file entries (so the skeleton entries do not cause jumping when the actual loaded content has less than 4 entries).

But if the actual listing has more than 4 entries the file picker size is "jumping". 
So to prevent this we must set a fixed height to the file picker, we could e.g. set it to 90vh¹.

cc @szaimen @jancborchardt @nimishavijay

¹ Not required for mobile as it is always full screen there.

susnux:
* Resolves #900 

This fixes some remaining styling issues and the second commit sets a fixed height to the file picker.
We already have a max. width of 900px and a max. height of 90% from the modal.

So from my perspective a height of max. 800px looks good (900x800px), thats why I set the height to `min(80vh, 800px)`.

small | medium | large
---|---|---
Screen Shot 2023-08-28 at 21 27 42|Screen Shot 2023-08-28 at 21 27 50|Screen Shot 2023-08-28 at 21 27 57

### DIFF
--- a/lib/components/DialogBase.vue
+++ b/lib/components/DialogBase.vue
@@ -5,14 +5,14 @@
 		@close="handleClose">
 		<!-- The dialog name / header -->
 		<h2 class="dialog__name" v-text="name" />
-		<div class="dialog">
+		<div class="dialog" :class="dialogClasses">
 			<div ref="wrapper" :class="['dialog__wrapper', { 'dialog__wrapper--collapsed': isNavigationCollapsed }]">
 				<!-- When the navigation is collapsed (too small dialog) it is displayed above the main content, otherwise on the inline start -->
-				<nav v-if="hasNavigation" :class="['dialog__navigation', ...navigationClasses]">
+				<nav v-if="hasNavigation" class="dialog__navigation" :class="navigationClasses">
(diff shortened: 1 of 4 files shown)


skjnldsv (approved):

jancborchardt (approved):
Looks good design-wise! But will this make it to 27.1 as mentioned in https://github.com/nextcloud-libraries/nextcloud-dialogs/issues/900#issuecomment-1680867330 @susnux @szaimen?

Edit: Ah I guess that's what the backport at https://github.com/nextcloud-libraries/nextcloud-dialogs/pull/972 is for? Let's just make sure it gets in.

susnux:
@jancborchardt If someone merges the open stable4 requests and creates a new beta release then we can try to merge on server until Thursday. 

@skjnldsv @AndyScherzinger

jancborchardt:
@AndyScherzinger @skjnldsv is that possible? Otherwise the filepicker changes height often which looks very off.

PR #973, opened 2023-08-28
Get files app config and use it for the file picker

### LINKED ISSUE #908: file picker should show hidden folders
Title should be explanatory enough.
Please see https://github.com/nextcloud/bookmarks/issues/2046

susnux:
* resolves #908 
* resolves https://github.com/nextcloud/server/issues/38926
  * resolves https://github.com/nextcloud/server/issues/39185

Done:
* [x] Get config
* [x] sort favorites first
* [x] show hidden files if configured
* [x] crop previews
* [x] sort by files app order
* [x] unit tests


lib/composables/filesSettings.ts
```diff
@@ -0,0 +1,125 @@
...
+			cropImagePreviews.value = respose.data?.data?.crop_image_previews ?? true
+		})
+	}
+
+	return {
+		showHiddenFiles,
+		sortFavoritesFirst,
+		cropImagePreviews,
+	}
+}
```
skjnldsv, line 74: Do you think we should have this in @nextcloud/files?
I feel like adding more config on server will make the upgrade/implementation process much more complicated :thinking:

susnux: Do you mean any function to access the config or the composable?
Generally I would say it would make sense to provide a common way to access the configuration.

skjnldsv: yes, but maybe we could have a way to not have those configs hardcoded somehow. So if we add a new one in the server backend, we don't have to update 20 other locations? :thinking: 
No idea how to pull that off :see_no_evil:

Pytal (approved):

PR #987, opened 2023-09-07
fix: breadcrumbs grow

skjnldsv:
Before/after
Peek 07-09-2023 16-30


skjnldsv:
> Where was it overridden?

No clue, might depend  on the loading order :thinking: 
Both are data-targeting attributes 
image

susnux:
> Where was it overridden?

Vue library

PR #1070, opened 2023-10-17
Fix package and add separate filepicker entry point

susnux:
1. Combine vite and vitest config
2. Remove non existing legacy entry point from package.json
3. Do not export the filepicker Vue component in main entry point. Instead export it in a `filepicker` entrypoint, also load the component async. in the filepicker builder. This ensures that if a user does not use the filepicker, it gets treeshaken.


Pytal (approved):

package.json
```diff
@@ -12,18 +12,14 @@
       "import": "./dist/index.mjs",
       "require": "./dist/index.cjs"
     },
-    "./dist/index.css": {
-      "import": "./dist/style.css",
-      "require": "./dist/style.css"
```
Pytal, line 17: Add that this is not needed in breaking changes @susnux ?

ref: https://github.com/nextcloud/server/blob/a21d6c5d8ddb58add6995dd4b0b2a885473f0677/core/css/server.scss#L11

PR #1095, opened 2023-10-31
FilePicker: Signal folder creation to files app

### LINKED ISSUE #1016: Emit `files:node:created` when folder is created
When creating a new directory in the filepicker we need to emit the eventbus event `files:node:created` to let the files app know (required if the file picker is used on the files app).

For context see https://github.com/nextcloud/server/issues/33323

susnux:
* Resolves #1016


skjnldsv (approved):
Seems like an easy way, but fetching the full list of file for a single created file isn't the best optimised way :)

susnux:
> Seems like an easy way, but fetching the full list of file for a single created file isn't the best optimised way :)

True, will do in a follow up

Issue #1119, opened 2023-11-21
Missing previews in file picker?


nickvergessen:
v5.0.2

![](https://user-images.githubusercontent.com/213943/284489964-dcd55b1a-4d55-4ee8-a4f4-df8780355bc5.png)

Also the padding on the left side is wrong.

susnux:
Regression of #1108

PR #1108, opened 2023-11-15
Do not load user config for each previews

artonge:

PR #1120, opened 2023-11-23
fix(FilePicker): `cropImagePreviews` is now provided as a prop in FilePreview

susnux:
* Regression of #1108
* Resolves: https://github.com/nextcloud-libraries/nextcloud-dialogs/issues/1119

We need to use the property instead of the previous available reference for `cropImagePreviews`.


Pytal (approved):

nickvergessen:
5.0.3 incoming for the final releases?

PR #1128, opened 2023-11-25
fix(build): include ?raw data to bundle

ShGKme:
Currently, importing raw SVG icons is externalized and used as it is in dist. However, such import is not valid and requires additional bundler configuration. Even when this library doesn't include CSS imports to the bundle.

### Before in `dist`

```js
import IconMove from '@mdi/svg/svg/folder-move.svg?raw'
import IconCopy from '@mdi/svg/svg/folder-multiple.svg?raw'
```

### After

```js
const IconMove = '<svg xmlns="http://www.w3.org/2000/svg" id="mdi-folder-move" viewBox="0 0 24 24"><path d="M14,18V15H10V11H14V8L19,13M20,6H12L10,4H4C2.89,4 2,4.89 2,6V18A2,2 0 0,0 4,20H20A2,2 0 0,0 22,18V8C22,6.89 21.1,6 20,6Z" /></svg>'
const IconCopy = '<svg xmlns="http://www.w3.org/2000/svg" id="mdi-folder-multiple" viewBox="0 0 24 24"><path d="M22,4H14L12,2H6A2,2 0 0,0 4,4V16A2,2 0 0,0 6,18H22A2,2 0 0,0 24,16V6A2,2 0 0,0 22,4M2,6H0V11H0V20A2,2 0 0,0 2,22H20V20H2V6Z" /></svg>'
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,6 @@
       "version": "5.0.2",
       "license": "GPL-3.0-or-later",
       "dependencies": {
-        "@mdi/svg": "^7.3.67",
         "@nextcloud/axios": "^2.4.0",
         "@nextcloud/event-bus": "^3.1.0",
         "@nextcloud/files": "^3.0.0",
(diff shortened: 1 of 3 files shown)


susnux:
In this case @mdi/svg can be move to devDependencies

pulsejet:
> In this case @mdi/svg can be move to devDependencies

What depends on it? Can't we just remove?

susnux (approved):

PR #1166, opened 2024-01-10
enh(FilePickerBuilder): Change error thrown when FilePicker is closed

susnux:
Change error thrown when FilePicker is closed to `FilePickerClosed` so users can distinguish if error or just closed.

### DIFF
--- a/lib/filepicker-builder.ts
+++ b/lib/filepicker-builder.ts
@@ -41,6 +41,11 @@ export enum FilePickerType {
 	Custom = 5,
 }
 
+/**
+ * Error that is thrown in the rejected promise when the FilePicker was closed
+ */
+export class FilePickerClosed extends Error {}
+
 export class FilePicker<IsMultiSelect extends boolean> {
 
 	private title: string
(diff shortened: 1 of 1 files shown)


skjnldsv (approved):

skjnldsv:
What about returning a cancelable promise?
And check if the promise is cancelled?

susnux:
That is already done, a promise is returned. But currently a generic Error is used for the rejection.
So you do no know if it is a real error or just aborted by user.

By using reject with a custom error you can check for the error instance to see if it was aborted by the user.

skjnldsv:
One could easily see it differently :see_no_evil: 
Cancel with resolve on a null value.
As the dev should check the sanity of the picked value afterwards anyway :shrug: 

In any case, I think a cancellation would also fit the use case like we do here: https://github.com/nextcloud/server/blob/c9fd5101fceab0d9887b12915435e10492b70674/apps/files/src/services/Files.ts#L80

susnux:
> Cancel with resolve on a null value.

Sounds good but breaking change. Lets do it for 6.x and NC 29 :)

PR #1169, opened 2024-01-12
feat(breadcrumb): added href and stopped event default on a element

emoral435:
## 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image|firefox_s9J1mvZeTU

## GIF For New Changes
firefox_tPXw1ltV1S

## 🚧 Summary
Added an HREF attribute to the link element. Once this gets merged / challenged / changed, will update the library in nextcloud/server

### DIFF
--- a/lib/components/FilePicker/FilePickerBreadcrumbs.vue
+++ b/lib/components/FilePicker/FilePickerBreadcrumbs.vue
@@ -3,7 +3,8 @@
 		<template #default>
 			<NcBreadcrumb :name="t('Home')"
 				:title="t('Home')"
-				@click="emit('update:path', '/')">
+				:href="getPath"
+				@click.prevent="emit('update:path', '/')">
 				<template #icon>
 					<IconHome :size="20" />
 				</template>
@@ -12,7 +13,8 @@
 				:key="dir.path"
 				:name="dir.name"
 				:title="dir.path"
-				@click="emit('update:path', dir.path)" />
+				:href="getPath"
+				@click.prevent="emit('update:path', dir.path)" />
 		</template>
 		<template v-if="showMenu" #actions>
 			<NcActions :aria-label="t('Create directory')"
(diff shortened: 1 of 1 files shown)


lib/components/FilePicker/FilePickerBreadcrumbs.vue
```diff
@@ -119,6 +121,11 @@ const pathElements = computed(() => props.path.split('/')
 		path: '/' + elements.slice(0, i + 1).join('/'),
 	})),
 )
+
+/**
+ * Gets the current url path
+ */
+const getPath = computed(() => window.location.pathname)
```
ShGKme, line 128: **Note:** `window.location.pathname` is not reactive. So `computed` won't recalculate on pathname change. If the location is not supposed to be changed, it can be a plain constant. If it can change, it should be a method (it is called as a method already).

ShGKme: But speaking about the issue, I don't think it should be a link at all. With this implementation, a user has a list of links, all to the current page. 

Probably, `NcBreadcrumb` should support rendering as a button when non `to` and `href` is provided.

cc @susnux and @JuliaKirschenheuter for a second opinion here.

susnux: I agree with your second comment, rendering it as a link does not make any sense in this case.

lib/components/FilePicker/FilePickerBreadcrumbs.vue
```diff
@@ -3,7 +3,8 @@
 		<template #default>
 			<NcBreadcrumb :name="t('Home')"
 				:title="t('Home')"
-				@click="emit('update:path', '/')">
+				:href="getPath"
```
susnux, line 6: I do not think this makes sense, this will be the current app which might be something other than files

lib/components/FilePicker/FilePickerBreadcrumbs.vue
emoral435: Yes, in this case, the way that these breadcrumbs are handled are that they each emit an event to reload the popup window, but it does not redirect the user to a different URL! So, I do agree, I think because the href in the breadcrumbs is essentially useless, we should change this to a list of buttons that have the same emit

emoral435:
Closing PR - listened to feedback, and decided that it would be better to render `<a/>` as `<button/>` instead! Better semantics

PR #1191, opened 2024-01-18
fix(FilePicker): Improve choose button text when selecting multiple

susnux:
Instead of only showing `Choose FIRST FILE` or just `Choose` show `Choose 3 files` when multiselect is enabled.

### DIFF
--- a/l10n/messages.pot
+++ b/l10n/messages.pot
@@ -20,6 +20,11 @@ msgstr ""
 msgid "Choose {file}"
 msgstr ""
 
+msgid "Choose %n file"
+msgid_plural "Choose %n files"
+msgstr[0] ""
+msgstr[1] ""
+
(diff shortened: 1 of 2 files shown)


come-nc (approved):

come-nc:
But it also change `Choose FIRST FILE` to `Choose 1 file`, no? Is that ok?

come-nc:
> But it also change `Choose FIRST FILE` to `Choose 1 file`, no? Is that ok?

Nope, my bad, it does not. It’s weird that the multiversions has plural and singular but I suppose it’s needed for other languages.

susnux:
> Nope, my bad, it does not. It’s weird that the multiversions has plural and singular but I suppose it’s needed for other languages.

yes there are languages with more plural forms

PR #1207, opened 2024-01-25
fix(FilePicker): Stop default close event in case of button press

susnux:
* Resolves https://github.com/nextcloud/server/issues/42469

If you press a button NcDialog will call the filepicker callback and afterwards close the dialog.
But our callback is async and NcDialog does not await the callback.

This means sometimes the close event with picked nodes is not emitted first but the close from NcDialog.

So what I do now:
1. On button press (now synchronous function) lock the close event
2. Call the callback
3. Emit only one close (that with the picked nodes)
4. Unlock the close event

### DIFF
--- a/lib/components/FilePicker/FilePicker.vue
+++ b/lib/components/FilePicker/FilePicker.vue
@@ -1,5 +1,5 @@
 <template>
-	<NcDialog v-bind="dialogProps" @update:open="handleClose">
+	<NcDialog v-bind="dialogProps" :open.sync="isOpen" @update:open="handleClose">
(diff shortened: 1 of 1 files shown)


Pytal (approved):
Would adding an awaited callback to the API make sense?

susnux:
> Would adding an awaited callback to the API make sense?

Yes but we need a solution for now. We can do later as otherwise we force every library user to upgrade nextcloud vue. Including stable27

ShGKme (approved):

skjnldsv (approved):

PR #1234, opened 2024-02-15
enh: breadcrumbs are now consistant across server

emoral435:
Coincides with nextcloud/server#43604
Instead of "Home", breadcrumb name is now "All files". Changed this to be more consistent across the server :)

PLACE TO CHANGE: Section that says "Home"
image

### DIFF
--- a/lib/components/FilePicker/FilePickerBreadcrumbs.vue
+++ b/lib/components/FilePicker/FilePickerBreadcrumbs.vue
@@ -1,7 +1,7 @@
 <template>
 	<NcBreadcrumbs class="file-picker__breadcrumbs">
 		<template #default>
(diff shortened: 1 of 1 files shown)


susnux:
But it is not all files but home or start or what ever you call the root?

emoral435:
I see what you mean! Both cases work well here, tbh. We can leave the regular name be "Home" (or root) here, as it does convey where the file picker is searching from, but IMO "All files" lets the user instantly know they are looking at all possible files, and is consistent design wise with the breadcrumbs that will be on `nextcloud/server`. If you still believe we can leave it how it is currently, then I can put this PR down :)

susnux (approved):
fine then by me

PR #1241, opened 2024-02-20
fix(FilePicker): Allow using on public shares

susnux:
* Resolves https://github.com/nextcloud/server/issues/42009

Only query settings from the files app when we are on the internal view.
Also adjusted the webdav endpoint used when we are on a public share.

### DIFF
--- a/l10n/messages.pot
+++ b/l10n/messages.pot
@@ -34,6 +34,12 @@ msgstr ""
 msgid "Could not create the new folder"
 msgstr ""
 
+msgid "Could not load files settings"
+msgstr ""
+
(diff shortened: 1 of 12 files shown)


emoral435 (approved):
Code LGTM! Haven't tested it, though!

ShGKme:
Is there a plan to have a release with this fix?

This allows using the component outside of the server rendered page, for example, in Talk Desktop.

susnux:
This is already released in 5.1.2

ShGKme:
> This is already released in 5.1.2

Thanks! Sorry, I checked only by milestone

PR #1284, opened 2024-03-29
fix(files): Fix newly created folder not appearing in main files view

Pytal:
A `files:node:created` event is emitted correctly but was missing the fileid

image

For #1285

### DIFF
--- a/lib/composables/dav.ts
+++ b/lib/composables/dav.ts
@@ -19,7 +19,7 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  *
  */
-import type { Folder, Node } from '@nextcloud/files'
+import { Folder, type Node } from '@nextcloud/files'
 import type { ComputedRef, Ref } from 'vue'
 import type { FileStat, ResponseDataDetailed, SearchResult } from 'webdav'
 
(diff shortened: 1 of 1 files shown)


lib/composables/dav.ts
```diff
@@ -85,8 +85,26 @@ export const useDAVFiles = function(
 	async function createDirectory(name: string): Promise<Folder> {
 		const path = join(currentPath.value, name)
 
-		await client.value.createDirectory(join(defaultRootPath.value, path))
-		const directory = await getFile(path) as Folder
+		const { headers } = await client.value.customRequest(
```
susnux, line 88: Why a custom request? The previously used `createDirectory` should work, if not then something else is broken.

lib/composables/dav.ts
```diff
@@ -85,8 +85,26 @@ export const useDAVFiles = function(
...
+			join(defaultRootPath.value, path),
+			{
+				method: 'MKCOL',
+			},
+		)
+		const stat = await getFile(path)
+		const directory = new Folder({
```
susnux, line 95: `getFile` already returns a Node so in this case a Folder. If this does not work then `@nextcloud/files` is broken

emoral435:
Updated comment to show redirection between chained PR's :)

lib/composables/dav.ts
Pytal: Used a custom request here to be able to access the headers as `createDirectory` doesn't return them

PR #1287, opened 2024-03-29
fix(FilePicker): Request all default file props (incl. file id)

susnux:
* Supersedes https://github.com/nextcloud-libraries/nextcloud-dialogs/pull/1284
* Supersedes https://github.com/nextcloud-libraries/nextcloud-dialogs/pull/1285
* Resolves https://github.com/nextcloud/server/issues/44482

The propfind was not requesting all needed props, this is fixed by using `davGetDefaultProps` which will then load the file id.


emoral435 (approved):

Pytal (approved):
Fixes it, thanks!

Issue #300, opened 2021-02-23
Include confirmDestructive Dialog


jotoeri:
Just wanted to use the confirmationDialog on Forms and saw i need to take it manually from `OC.dialogs`. Would imo be nice to have it included in this package!

susnux:
This will not make it into 4.2.0 but I think about implementing the dialogs based on the new dialog component for 5.0.0, but we could also backport this for 4.3.0 to support this on Nextcloud 27 as well.

PR #1297, opened 2024-04-09
feat: Provide generic dialogs previously provided by `OC.dialogs`

susnux:
* Resolves: https://github.com/nextcloud-libraries/nextcloud-dialogs/issues/300 

Lets clean up legacy jQuery UI and move to Vue dialogs.

### DIFF
--- a/lib/components/GenericDialog.vue
+++ b/lib/components/GenericDialog.vue
@@ -1,0 +1,66 @@
+<template>
+	<NcDialog dialog-classes="nc-generic-dialog"
+		:buttons="buttons"
+		:name="name"
+		:message="text"
+		@update:open="$emit('close')">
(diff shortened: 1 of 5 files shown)


lib/components/GenericDialog.vue
```diff
@@ -0,0 +1,66 @@
...
+	 */
+	html?: string
+
+	/**
+	 * Buttons on the dialog
+	 */
+	buttons?: IDialogButton[]
+
+	/**
+	 * Severity of the dialog - if se a notecard is used
```
Pytal, line 46:
```suggestion
	 * Severity of the dialog - if a notecard is used
```

Pytal (approved):
Code looks good!

PR #1347, opened 2024-06-04
[stable4] fix(deps): bump @nextcloud/files to 3.0.0

st3iny:
The required `@nextcloud/files` dependency is broken and fixing this inside other apps is hacky because the package lock has to be adjusted manually. It would be nice to release a 4.x.x version with the fixed dependency to make it easier to fix downstream.

Note that simply adding `@nextcloud/files` as a dependency to an app's package.json does not work because it is not a peer of `@nextcloud/dialogs`. Thus, npm will just install both versions separately. 

E.g. in Deck: 
```
@nextcloud/files@3.0.0-beta.21
node_modules/@nextcloud/dialogs/node_modules/@nextcloud/files
  @nextcloud/files@"3.0.0-beta.21" from @nextcloud/dialogs@4.2.7
  node_modules/@nextcloud/dialogs
    @nextcloud/dialogs@"^4.2.2" from the root project

@nextcloud/files@3.2.1
node_modules/@nextcloud/files
  @nextcloud/files@"^3.0.0" from the root project
```

Ref
- https://github.com/nextcloud/deck/issues/5935
- https://github.com/nextcloud/deck/issues/5822
- https://github.com/nextcloud/calendar/issues/6018

Exemplary (hacky) downstream fix at https://github.com/nextcloud/calendar/pull/6035

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "GPL-3.0-or-later",
       "dependencies": {
         "@mdi/svg": "^7.4.47",
-        "@nextcloud/files": "3.0.0-beta.21",
+        "@nextcloud/files": "3.0.0",
         "@nextcloud/l10n": "^2.2.0",
         "@nextcloud/router": "^2.2.0",
         "@nextcloud/typings": "^1.7.0",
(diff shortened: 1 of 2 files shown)


susnux:
No this will break server on Nextcloud 27!

stable4 is only intended to be used with Nextcloud <= 27.
If you need to support a wider range please use v5 which also will work with Nextcloud 28+ and probably even 27.

susnux:
See also:
- https://github.com/nextcloud-libraries/nextcloud-dialogs/pull/1226
- https://github.com/nextcloud/server/pull/43368

PR #1351, opened 2024-06-08
fix: Also add current folder to button factory

susnux:
When picking directories and no node is selected the current directory is picked, this is correctly submitted but it is currently not correctly included in the button factory.

Meaning you can for example not check if you can pick the current directory (readonly directory).

Required for https://github.com/nextcloud/server/pull/45688

### DIFF
--- a/lib/components/FilePicker/FilePicker.vue
+++ b/lib/components/FilePicker/FilePicker.vue
@@ -18,9 +18,10 @@
 		<div class="file-picker__main">
 			<!-- Header title / file list breadcrumbs -->
 			<FilePickerBreadcrumbs v-if="currentView === 'files'"
-				:path.sync="currentPath"
+				:path="currentPath"
 				:show-menu="allowPickDirectory"
-				@create-node="onCreateFolder" />
+				@create-node="onCreateFolder"
+				@update:path="navigatedPath = $event" />
 			<div v-else class="file-picker__view">
 				<h3>{{ viewHeadline }}</h3>
 			</div>
@@ -67,7 +68,7 @@ import FilePickerNavigation from './FilePickerNavigation.vue'
 
 import { emit as emitOnEventBus } from '@nextcloud/event-bus'
 import { NcDialog, NcEmptyContent } from '@nextcloud/vue'
-import { computed, onMounted, ref, toRef } from 'vue'
+import { computed, onMounted, ref, shallowReactive, toRef, watch } from 'vue'
 import { showError } from '../../toast'
 import { useDAVFiles } from '../../composables/dav'
 import { useMimeFilter } from '../../composables/mime'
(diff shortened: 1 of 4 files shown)


lib/components/FilePicker/FilePicker.vue
```diff
@@ -189,7 +190,7 @@ const viewHeadline = computed(() => currentView.value === 'favorites' ? t('Favor
 /**
  * All currently selected files
  */
-const selectedFiles = ref<Node[]>([])
+const selectedFiles = shallowReactive<Node[]>([])
```
susnux, line 193: This needs to be a shallow reactive, meaning the *array* is reactive, but not the inner elements.
This is required to fix Typescript, as Vue unrefs the inner nodes for `ref` which will strip off any private properties of the Node type. This causes Typescript issues as the types `Ref<Node>.value != Node`.

lib/components/FilePicker/FilePicker.vue
```diff
@@ -200,28 +201,23 @@ const savedPath = ref(window?.sessionStorage.getItem('NC.FilePicker.LastPath') |
 /**
  * The path the user manually navigated to using this filepicker instance
  */
-const navigatedPath = ref<string>()
+const navigatedPath = ref('')
+// Save the navigated path to the session storage on change
+watch([navigatedPath], () => {
```
susnux, line 206: Instead of the writable computed below we can simply watch this and then react to it. Because doing login in writable computed properties seems to be bad.

lib/composables/dav.ts
```diff
@@ -114,7 +114,15 @@ export const useDAVFiles = function(
 	/**
 	 * All files in current view and path
 	 */
-	const files = ref<Node[]>([] as Node[]) as Ref<Node[]>
+	const files = shallowRef<Node[]>([] as Node[]) as Ref<Node[]>
```
susnux, line 117: Same reason as for the `shallowReactive` but here we can use `shallowRef` because it only needs to be reactive on change of the value, we do not change the array.

lib/components/types.ts
```diff
@@ -46,6 +46,12 @@ export interface IDialogButton {
 	 * @see https://nextcloud-vue-components.netlify.app/#/Components/NcButton
 	 */
 	type?: 'primary' | 'secondary' | 'error' | 'warning' | 'success'
+
+	/**
+	 * Disabled state of the button
+	 * @default false
+	 */
+	disabled?: boolean
```
artonge, line 54: I assume that `setButtonFactory` is called on every navigation change, so the disable state can be updated accordingly?

artonge (approved):

lib/components/types.ts
susnux: Yes exactly

ShGKme (approved):

PR #1354, opened 2024-06-12
Add SPDX header

AndyScherzinger:
While adding SPDX headers I realized the whole source had AGPL3+ license headers, hence I updated the LICENSE and package.json files to reflect that, effectively replacing the GPL3+ text with AGPL3+.

Not sure if this will be automatically reflected on https://www.npmjs.com/package/@nextcloud/dialogs

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,5 +1,5 @@
-                    GNU GENERAL PUBLIC LICENSE
-                       Version 3, 29 June 2007
+                    GNU AFFERO GENERAL PUBLIC LICENSE
+                       Version 3, 19 November 2007
 
  Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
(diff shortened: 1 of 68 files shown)


LICENSE
```diff
@@ -1,23 +1,21 @@
-                    GNU GENERAL PUBLIC LICENSE
-                       Version 3, 29 June 2007
+                    GNU AFFERO GENERAL PUBLIC LICENSE
```
susnux, line 1: This needs to be fixed in the `package.json` as well then.

AndyScherzinger: Fixed via https://github.com/nextcloud-libraries/nextcloud-dialogs/pull/1354/commits/3f26d5cc7732ef0341fd871f1b935f9520791472

skjnldsv (approved):

PR #1363, opened 2024-06-21
fix: Prevent invalid range for skeleton number

susnux:
Vue can not handle `v-for="index in -1"` so ensure the number is positive.

### DIFF
--- a/lib/components/FilePicker/FileList.vue
+++ b/lib/components/FilePicker/FileList.vue
@@ -201,7 +201,7 @@ const fileContainer = ref<HTMLDivElement>()
 			}
 		}
 		// container height - 50px table header / row height of 50px
-		skeletonNumber.value = Math.floor((height - 50) / 50)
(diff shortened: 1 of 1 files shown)


lib/components/FilePicker/FileList.vue
```diff
@@ -56,7 +56,7 @@
 					<LoadingTableRow v-for="index in skeletonNumber" :key="index" :show-checkbox="multiselect" />
 				</template>
 				<template v-else>
-					<FileListRow v-for="file in sortedFiles"
+					<FileListRow v-for="file of sortedFiles"
```
ShGKme, line 59: I think, it can confuse, making a feeling it is something different

ShGKme (approved):

lib/components/FilePicker/FileList.vue
susnux: Reverted this as not really related, but I would say it is the opposite.
`for ... of ...` is the JS loop for iterators, while `for ... in ...` mostly does not do what people expect it to do

PR #1378, opened 2024-07-02
fix(FilePicker): Cleanup DAV handling and properly handle `currentFolder`

susnux:
Instead of fetching the directory in a second request we simply do it like the file app and always request the content together with its root. This also allows to remove some functions by refactoring.


susnux:
Will make Cypress on server less flaky

artonge (approved):

Issue #1387, opened 2024-07-09
Requesting file previews should be limited in concurrency


SystemKeeper:
Currently we request all previews for the files in a folder at once

https://github.com/nextcloud-libraries/nextcloud-dialogs/blob/3e49a75698d930c39beeec12f31c84a172fe40e1/lib/components/FilePicker/FilePreview.vue#L38-L47

This will be problematic for folders with a large amount of files and/or previews that are not already cached. We should therefore limited the amount of requests we do in parallel as it is done on file deletion with a `PQueue` for example:

https://github.com/nextcloud/server/blob/025a7849b487351d0240d89833b3ab825897097d/apps/files/src/actions/deleteAction.ts#L106

Alternatively the previews should only be requested if they are close to "getting into the visible view"?!

PR #1478, opened 2024-10-29
feat: Rate-limit image previews

susnux:
* Resolves https://github.com/nextcloud-libraries/nextcloud-dialogs/issues/1387

Image preloading should be rate limited, to not overload the server. Now it is set to max. 5 requests at the same time.

### DIFF
--- a/lib/utils/imagePreload.ts
+++ b/lib/utils/imagePreload.ts
@@ -1,0 +1,25 @@
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import PQueue from 'p-queue'
+
+const queue = new PQueue({ concurrency: 5 })
+
+/**
+ * Preload an image URL
+ * @param url URL of the image
+ */
+export function preloadImage(url: string): Promise<boolean> {
+	const { resolve, promise } = Promise.withResolvers<boolean>()
+	queue.add(() => {
+		const image = new Image()
+		image.onerror = () => resolve(false)
+		image.onload = () => resolve(true)
+		image.src = url
+		return promise
+	})
+
+	return promise
+}
--- a/lib/components/FilePicker/FilePreview.vue
+++ b/lib/components/FilePicker/FilePreview.vue
@@ -3,9 +3,9 @@
   - SPDX-License-Identifier: AGPL-3.0-or-later
 -->
 <template>
-	<div :style="canLoadPreview ? { backgroundImage: `url(${previewURL})`} : undefined"
+	<div :style="previewLoaded ? { backgroundImage: `url(${previewURL})`} : undefined"
 		:class="fileListIconStyles['file-picker__file-icon']">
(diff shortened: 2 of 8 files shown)


lib/utils/imagePreload.ts
```diff
@@ -0,0 +1,24 @@
...
+ * @param url URL of the image
+ */
+export function preloadImage(url: string): Promise<boolean> {
+	const { resolve, promise } = Promise.withResolvers<boolean>()
+	queue.add(() => {
+		const image = new Image()
+		image.onerror = () => resolve(false)
+		image.onload = () => resolve(true)
+		image.src = url
+	})
```
Antreesy, line 21: queued callback itself is not waiting for preview to be fetched, so they're executed all at once, killing the efforts to rate-limit them =)

Works for me with adding one more layer of Promises (compare Network waterfall in both cases):
```suggestion
	queue.add(async () => {
		const image = new Image()
		const imagePromise = new Promise<boolean>((imgResolve) => {
			image.onload = () => imgResolve(true)
			image.onerror = () => imgResolve(false)
		})
		image.src = url
		const result = await imagePromise
		resolve(result)
	})
```

Antreesy (commented):
Everything apart from rate-limiting looks and works fine 🙈

lib/utils/imagePreload.ts
Antreesy: Shall we also clear the queue on FilePicker unmount?

susnux:
```suggestion
	queue.add(() => {
		const image = new Image()
		image.onerror = () => resolve(false)
		image.onload = () => resolve(true)
		image.src = url
		return promise
	})
```

susnux:
> Everything apart from rate-limiting looks and works fine 🙈

Should be fixed, somehow forgot to return the promise

Antreesy (approved):

lib/utils/imagePreload.ts
```diff
@@ -0,0 +1,25 @@
...
+import PQueue from 'p-queue'
+
+const queue = new PQueue({ concurrency: 5 })
+
+/**
+ * Preload an image URL
+ * @param url URL of the image
+ */
+export function preloadImage(url: string): Promise<boolean> {
+	const { resolve, promise } = Promise.withResolvers<boolean>()
```
Antreesy, line 15: https://caniuse.com/?search=promise.withresolvers

Should be polyfilled for now, until it has normal support?
```js
// @ts-expect-error This does not exist outside of polyfill which this is doing
if (typeof Promise.withResolvers === 'undefined') {
    if (window)
        // @ts-expect-error This does not exist outside of polyfill which this is doing
        window.Promise.withResolvers = function () {
            let resolve, reject
            const promise = new Promise((res, rej) => {
                resolve = res
                reject = rej
            })
            return { promise, resolve, reject }
        };
}
```

susnux: > it has normal support?

It already has, supported by every browser, also supported by Node - but we need to upgrade to LTS Node version 22.

Also: We already polyfill this for Nextcloud server, so the only problem here are the tests.

PR #1545, opened 2025-01-15
fix(files): Add encrypted icon for end-to-end encrypted folder

JuliaKirschenheuter:
|  Before   | After                                                        |
|-------------|---------------------------------------------------------------|
| Screenshot from 2025-01-15 12-48-48 | Screenshot from 2025-01-15 12-44-49

### DIFF
--- a/lib/components/FilePicker/FilePreview.vue
+++ b/lib/components/FilePicker/FilePreview.vue
@@ -7,6 +7,7 @@
 		:class="fileListIconStyles['file-picker__file-icon']">
 		<template v-if="!previewLoaded">
 			<IconFile v-if="isFile" :size="20" />
+			<component :is="folderDecorationIcon" v-else-if="folderDecorationIcon"/>
 			<IconFolder v-else :size="20" />
 		</template>
 	</div>
@@ -19,6 +20,12 @@ import { usePreviewURL } from '../../composables/preview'
 
 import IconFile from 'vue-material-design-icons/File.vue'
 import IconFolder from 'vue-material-design-icons/Folder.vue'
+import LockIcon from 'vue-material-design-icons/Lock.vue'
+import TagIcon from 'vue-material-design-icons/Tag.vue'
+import LinkIcon from 'vue-material-design-icons/Link.vue'
+import AccountPlusIcon from 'vue-material-design-icons/AccountPlus.vue'
+import NetworkIcon from 'vue-material-design-icons/Network.vue'
+import AccountGroupIcon from 'vue-material-design-icons/AccountGroup.vue'
 
 // CSS modules
 import fileListIconStylesModule from './FileListIcon.module.scss'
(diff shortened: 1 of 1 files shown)


lib/components/FilePicker/FilePreview.vue
```diff
@@ -37,6 +39,14 @@ const {
 } = usePreviewURL(toRef(props, 'node'), computed(() => ({ cropPreview: props.cropImagePreviews })))
 
 const isFile = computed(() => props.node.type === FileType.File)
+const folderDecorationIcon = computed(() => {
+	if (props.node.type !== FileType.Folder) {
+		return null
+	}
+	if (props.node.attributes?.['is-encrypted'] === 1) {
+		return KeyIcon
+	}
```
juliusknorr, line 48: Maybe worth to also implement the others for gorupfolders/external storages right away as well like in the files app? https://github.com/nextcloud/server/blob/7c59119c03c491f3534021b7011adfebf6e27b19/apps/files/src/components/FileEntry/FileEntryPreview.vue#L188-L227

lib/components/FilePicker/FilePreview.vue
```diff
@@ -7,6 +7,7 @@
 		:class="fileListIconStyles['file-picker__file-icon']">
 		<template v-if="!previewLoaded">
 			<IconFile v-if="isFile" :size="20" />
+			<component :is="folderDecorationIcon" v-else-if="folderDecorationIcon"/>
```
juliusknorr, line 10: In files we have a folder icon with the decorative as an overlay, but I can see how this does not work well here UI wise with the lines being to small. 

@susnux @marcoambrosini Any further input here or would you be fine with the plain icon?

susnux: I would use the plain icon for now and later we should move to same size on file picker and file list. But this can wait IMHO, so stick with @JuliaKirschenheuter version :)

lib/components/FilePicker/FilePreview.vue
susnux: Agree.

lib/components/FilePicker/FilePreview.vue
```diff
@@ -19,6 +20,7 @@ import { usePreviewURL } from '../../composables/preview'
 
 import IconFile from 'vue-material-design-icons/File.vue'
 import IconFolder from 'vue-material-design-icons/Folder.vue'
+import KeyIcon from 'vue-material-design-icons/Key.vue'
```
susnux, line 23: Not sure dont we use the lock icon instead on files?

lib/components/FilePicker/FilePreview.vue
JuliaKirschenheuter: Done besides CollectivesIcon as is not available

susnux (approved):

lib/components/FilePicker/FilePreview.vue
AndyScherzinger: Yes, we use the lock icon

AndyScherzinger (approved):

juliusknorr (approved):

skjnldsv:
@JuliaKirschenheuter adding stuff is considered a feature.
Please make sure to use `feat` next time :)

PR #1606, opened 2025-02-28
chore: simplify vitest config and fix some build issues

susnux:
Fixed issues:
- When building with `dev` we were not enabling vue-devtools. This is fixed by setting `NODE_ENV=development`.
- The tsconfig rootdir was not aligned with the `includes` section. All included files need to be within the root dir. So we now have one config for production code and one for setup and tests.
- The config option "rollup types" will create independent type, this causes the filepicker module to be incompatible with the index module we exported. Fixed by not rolling up types (note: roll up types only works when there is only a single entry point)

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -1,0 +1,1 @@
+NODE_ENV=development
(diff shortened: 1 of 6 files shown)


.env.development
ShGKme, line 1: I was sure there is a way to set NODE_ENV in Vite via arg... But probably it was only available in Webpack.

Not a fun of having `.env.development` file just to enable devtools in dev, but it looks like it is the best cross-platform solution if we want to build on Windows as well.

ShGKme (approved):

PR #1609, opened 2025-02-28
chore: update workflows

susnux:
Should fix most of the security issues: https://github.com/nextcloud-libraries/nextcloud-dialogs/security


szaimen (approved):
🐘

PR #1690, opened 2025-04-01
fix: minor issues in the test suite

lacrioque:
Fixed the tests. 

## FileListRow.spec.ts
- Updated the syntax to not check against the non-existant .isEmpty, but to see if there are four `td` elements rendered

## dav.ts
- There was no fallback check if the content returned is completely  null for some reason or the other.
- A better test would be to mock the returns of the content variable somehow, but this should suffice for a general workflow


.gitignore
```diff
@@ -62,4 +62,3 @@ typings/
 # next.js build output
 .next
 
-/dist
```
susnux, line 65: Should not be committed!

bun.lock
susnux, line 1: should not be committed, we use node + npm

susnux (changes requested):
please do not commit dist files

PR #1734, opened 2025-04-29
perf: do not pre-load folder mime icon

susnux:
They will all fail with 404 and spam the server.
So we already use folder icons and do not use the result anyways.


Antreesy (approved):
Looks valid

Issue #1126, opened 2023-11-24
Dependencies on node APIs


pulsejet:
Can we _please_ not depend on `path`? I would really like to get rid of node polyfills.

susnux:
What is the issue you face? Because bundling those polyfills does not make much sense.
You will always need a bundler, so you can inject the polyfills your self, which will allow treeshaking and smaller bundle size as otherwise the same functions might get included multiple times just because they are bundled with each library.

pulsejet:
Actually none of the other nextcloud packages (at least the ones I used) depend on node polyfills, which allowed me to get rid of `NodePolyfillPlugin` altogether. But then I had to add it back just for this package. If we use something like `path-browserify` instead that actually targets the browser then there's no need to change the downstream application's webpack config.

pulsejet:
Also noteworthy: `path` depends on `process` so this also adds completely redundant code to the output bundle.

pulsejet:
Hmm, the docs of `path-browserify` claim webpack injects it by default. I'm a bit confused now.

pulsejet:
> Meaning for this library if we include the path polyfills you still need the polyfills in your app config because this library depends on e.g. `@nextcloud/files` which makes heavy use of the `path` module.

Gotcha, didn't know we use it in files too.

> So I would prefer if we can agree on a common practice across `@nextcloud/` libraries for handling node modules.

Yeah this makes total sense. From my side, the reasons I think we should move away from node polyfilling:
1. This code is only ever supposed run in the browser, so polyfilling node functionailty is semantically not right. This code is _never_ supposed to run in node, so it makes no sense to use node APIs.
2. It adds transitive dependencies that we don't need (e.g. process)
3. (most important) this causes the library to depend on the webpack configuration of the downstream plugin. Further, no idea what happens if the downstream app wants to switch to something other than webpack. The libraries should be as independent as possible.

Need input from more people on this. @skjnldsv any comments?

BTW the webdav client used to depend on node polyfills earlier, but that dependency too is gone in the latest version (they split node and browser modules).

skjnldsv:
> Need input from more people on this. @skjnldsv any comments?

Always up for cleaner approaches. Thought what library do you have in mind for the few path imports we're using (usually `normalize`, `basename`, `extname`, `dirname` and `join`)

pulsejet:
[...]

PR #1752, opened 2025-05-02
refactor: do not use Node internals but `@nextcloud/paths` package

susnux:
* Resolved https://github.com/nextcloud-libraries/nextcloud-dialogs/issues/1126


pulsejet (approved):
Thanks for this 🙏

PR #1759, opened 2025-05-07
[stable6] feat: add guest auth prompt component

skjnldsv:
- [x] Needs https://github.com/nextcloud-libraries/nextcloud-auth/pull/795 (to bump afterwards)

### DIFF
--- a/lib/components/PublicAuthPrompt.vue
+++ b/lib/components/PublicAuthPrompt.vue
@@ -1,0 +1,223 @@
+<!--
+  - SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<template>
+	<NcDialog :buttons="dialogButtons"
+		class="public-auth-prompt"
+		data-cy-public-auth-prompt-dialog
+		is-form
+		no-close
+		:name="title"
+		@submit="onSubmit">
+		<p v-if="text" class="public-auth-prompt__text">
+			{{ text }}
+		</p>
+
+		<!-- Header -->
+		<NcNoteCard class="public-auth-prompt__header"
+			:text="notice"
+			type="info" />
+
+		<!-- Form -->
+		<NcTextField ref="input"
+			class="public-auth-prompt__input"
+			data-cy-public-auth-prompt-dialog-name
+			:label="t('Name')"
+			:placeholder="t('Enter your name')"
+			:required="!cancellable"
+			v-model="name"
+			minlength="2"
+			name="name" />
+	</NcDialog>
+</template>
+
+<script lang="ts">
+import { defineComponent } from 'vue'
+import { getBuilder } from '@nextcloud/browser-storage'
+import { setGuestNickname } from '@nextcloud/auth'
+import { showError } from '@nextcloud/dialogs'
+
+import NcButton from '@nextcloud/vue/components/NcButton'
+import NcDialog from '@nextcloud/vue/components/NcDialog'
+import NcNoteCard from '@nextcloud/vue/components/NcNoteCard'
+import NcTextField from '@nextcloud/vue/components/NcTextField'
+
+import { t } from '../utils/l10n.ts'
+
+const storage = getBuilder('public').build()
+
+export default defineComponent({
+	name: 'PublicAuthPrompt',
+
+	components: {
+		NcDialog,
+		NcNoteCard,
+		NcTextField,
+	},
+
+	props: {
+		/**
+		 * Preselected nickname
+		 * @default '' No name preselected by default
+		 */
+		nickname: {
+			type: String,
+			default: '',
+		},
+
+		/**
+		 * Dialog title
+		 */
+		title: {
+			type: String,
+			default: t('Guest identification'),
+		},
+
+		/**
+		 * Dialog text under the dialog title
+		 * e.g 'Enter your name to access the file'
+		 * @default '' Not shown by default
+		 */
+		 text: {
+			type: String,
+			default: '',
+		},
+
+		/**
+		 * Dialog notice
+		 * @default 'You are currently not identified.'
+		 */
+		notice: {
+			type: String,
+			default: t('You are currently not identified.'),
+		},
+
+		/**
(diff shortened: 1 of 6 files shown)


lib/public-auth.ts
```diff
@@ -0,0 +1,28 @@
...
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import type { ComponentProps } from 'vue-component-type-helpers'
+import { defineAsyncComponent } from 'vue'
+import { spawnDialog } from '@nextcloud/vue/functions/dialog'
+
+import type PublicAuthPrompt from './components/PublicAuthPrompt.vue'
+
+type PublicAuthPromptProps = ComponentProps<typeof PublicAuthPrompt>
```
susnux, line 12: IMHO define the props type here expert them as an interface and then just use them for the component like `defineProps<PublicAuthPromptProps>()`

skjnldsv: Is there an advantage/inconvenient to it ?

skjnldsv: Because that seems far less clean that just using the vue-component-type-helpers :see_no_evil:

lib/components/PublicAuthPrompt.vue
```diff
@@ -0,0 +1,223 @@
...
+		},
+
+		/**
+		 * Dialog notice
+		 * @default 'You are currently not identified.'
+		 */
+		notice: {
+			type: String,
+			default: t('You are currently not identified.'),
+		},
```
susnux, line 97: Why should you need to customize this? This should be automatically calculated based on the `getGuestUser` output?

lib/components/PublicAuthPrompt.vue
```diff
@@ -0,0 +1,223 @@
...
+
+		/**
+		 * Dialog submit button label
+		 * @default 'Submit name'
+		 */
+		submitLabel: {
+			type: String,
+			default: t('Submit name'),
+		},
```
susnux, line 106: Same here, this seems to make no sense to customize and should rather be a sane default

lib/components/PublicAuthPrompt.vue
```diff
@@ -0,0 +1,223 @@
...
+
+		/**
+		 * Dialog subtitle
+		 * @default 'Enter your name to access the file'
+		 */
+		subtitle: {
+			type: String,
+			default: '',
+		},
```
susnux, line 88: This is not really a subtitle but its the dialog text (its a `<p>` not a `<h#>`), we also call it *text* on the other dialogs, so would make sense to keep it consistent.

Also not sure if customization is needed (could also be generic "resource") but maybe makes sense :)

lib/components/PublicAuthPrompt.vue
skjnldsv: Same answer: `Submit name and join`
image

lib/components/PublicAuthPrompt.vue
skjnldsv: Done

lib/components/PublicAuthPrompt.vue
susnux: I see, makes sense :)

lib/components/PublicAuthPrompt.vue
```diff
@@ -0,0 +1,224 @@
...
+  - SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<template>
+	<NcDialog :buttons="dialogButtons"
+		class="public-auth-prompt"
+		data-cy-public-auth-prompt-dialog
+		is-form
+		:can-close="false"
```
susnux, line 11: deprecated and removed in v9

```suggestion
		no-close
```

lib/components/PublicAuthPrompt.vue
```diff
@@ -0,0 +1,224 @@
...
+			type="info" />
+
+		<!-- Form -->
+		<NcTextField ref="input"
+			class="public-auth-prompt__input"
+			data-cy-public-auth-prompt-dialog-name
+			:label="t('Name')"
+			:placeholder="t('Enter your name')"
+			:required="!cancellable"
+			:value.sync="name"
```
susnux, line 30: deprecated and removed in v9

```suggestion
			v-model="name"
```

lib/components/PublicAuthPrompt.vue
```diff
@@ -0,0 +1,224 @@
...
+import NcNoteCard from '@nextcloud/vue/components/NcNoteCard'
+import NcTextField from '@nextcloud/vue/components/NcTextField'
+
+import { t } from '../utils/l10n.ts'
+
+const storage = getBuilder('public').build()
+
+// TODO: Remove this when the type export is fixed in the vue components library
+type ButtonType = InstanceType<typeof NcButton>['$props']['type']
+
```
susnux, line 53: not needed / deprecated / wrong type (should be by the way the type from NcDialogButton)

```suggestion
```

lib/components/PublicAuthPrompt.vue
```diff
@@ -0,0 +1,224 @@
...
+		return {
+			name: '',
+		}
+	},
+
+	computed: {
+		dialogButtons() {
+			const cancelButton = {
+				label: t('Cancel'),
+				type: 'tertiary' as ButtonType,
```
susnux, line 135: deprecated and removed in v9

```suggestion
				variant: 'tertiary',
```

lib/components/PublicAuthPrompt.vue
```diff
@@ -0,0 +1,224 @@
...
+				type: 'primary' as ButtonType,
+				nativeType: 'submit',
+			}
+
+			// If the dialog is cancellable, add a cancel button
+			if (this.cancellable) {
+				return [cancelButton, submitButton]
+			}
+
+			 return [submitButton]
```
susnux, line 150:
```suggestion
			return [submitButton]
```

skjnldsv:
I'm on mobile rn, will squash and adjust in a bit :)

skjnldsv:
Done!

susnux (approved):

PR #1774, opened 2025-05-11
refactor: adjust code to comply with code style (again)

susnux:
Noticed we do not have a linting CI here so we did not notice we merged stuff not complying with ESLint.
So this does:
- Fix code style and logic issues where needed
- Add CI job for ESLint

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -12,6 +12,12 @@ export {
 } from './filepicker-builder.ts'
 
 export {
+	type GuestUserPromptOptions,
+
+	showGuestUserPrompt,
+} from './public-auth.ts'
+
+export {
 	showError,
 	showInfo,
 	showLoading,
--- a/lib/public-auth.ts
+++ b/lib/public-auth.ts
@@ -19,9 +19,10 @@ export type GuestUserPromptOptions = PublicAuthPromptProps
  * @return The selected name or undefined if dialog was closed
  */
 export async function showGuestUserPrompt(props: GuestUserPromptOptions): Promise<string | undefined> {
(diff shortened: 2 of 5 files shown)


lib/public-auth.ts
```diff
@@ -5,21 +5,23 @@
...
-export function showGuestUserPrompt(props: PublicAuthPromptProps) {
-	return new Promise((resolve) => {
-		spawnDialog(
-			defineAsyncComponent(() => import('./components/PublicAuthPrompt.vue')),
-			props,
-			resolve,
-		)
-	})
+export async function showGuestUserPrompt(props: GuestUserPromptOptions): Promise<string | undefined> {
+	const [name] = await spawnDialog(
```
ShGKme, line 22: Tested? `spawnDialog` was supposed to return value as it is when possible

```suggestion
	const name = await spawnDialog(
```

lib/index.ts
```diff
@@ -11,6 +11,12 @@ export {
 	getFilePickerBuilder,
 } from './filepicker-builder.ts'
 
+export {
+	type GuestUserPromptOptions,
+
+	showGuestUserPrompt,
+} from './public-auth.ts'
+
```
ShGKme, line 19: Does it mean that a new function is now available in the library?

skjnldsv: Yes, merged yesterday

lib/public-auth.ts
susnux: Types are broken. Fixed the code here until its fixed in library.

skjnldsv (approved):

Issue #1773, opened 2025-05-11
pick() always will return an array even for multiselect === true even if the TypeScript typing is tuned to claim this different.


rotdrop:
In my NC instance the calendar attachments do not work, as this uses the file-picker in single-select mode, however, looking at the following line of code in the dialogs package

https://github.com/nextcloud-libraries/nextcloud-dialogs/blob/961a4d07345459292b16c27239ad8b284681ce67/lib/filepicker-builder.ts#L103

reveals that pick() always will return an array of strings (JS Array.map() will return an array, there is nothing to argue about), but the TS types in that line are tuned to lie about this fact.

The return value in the transpiled code will always be an array. So either the TS type-hints should be adjusted accordingly, or the code should be fixed to act as intended.

susnux:
Ah yes seems like a regression of 6.2.0

PR #1775, opened 2025-05-11
fix(FilePickerBuilder): correctly return array / plain value depending on multiselect

susnux:
* Resolves https://github.com/nextcloud-libraries/nextcloud-dialogs/issues/1773

### DIFF
--- a/lib/filepicker-builder.ts
+++ b/lib/filepicker-builder.ts
@@ -98,9 +98,10 @@ export class FilePicker<IsMultiSelect extends boolean> {
 	public async pick(): Promise<IsMultiSelect extends true ? string[] : string> {
 		const nodes = await this.pickNodes()
 		if (this.multiSelect) {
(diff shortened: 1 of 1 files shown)


rotdrop:
The changed line 101 contains a duplicated  check for "multiselect". I suppose this is not intentional?

susnux:
> The changed line 101 contains a duplicated check for "multiselect". I suppose this is not intentional?

Not sure what you mean?

rotdrop:
> > The changed line 101 contains a duplicated check for "multiselect". I suppose this is not intentional?
> 
> Not sure what you mean?

Line 101 after applying your commit:
```
return nodes.map((node) => node.path) as (IsMultiSelect extends true ? string[] : string)
```
This is already in the code path where multiselect is true. Ok. I suppose this is still needed in order to persuade TypeScript to deduce the correct return type ...

skjnldsv (approved):

PR #1836, opened 2025-06-10
fix: display guest name validity

skjnldsv:


susnux (approved):

lib/components/PublicAuthPrompt.vue
```diff
@@ -70,6 +71,18 @@ watch(() => props.nickname, () => {
 	name.value = props.nickname
 })
 
+watch(() => name.value, (newName) => {
```
skjnldsv, line 74: @susnux I'm not sure here, should it be this instead?

```suggestion
watch(name, (newName) => {
```

susnux: Only needed for `props` as the prop itself is not reactive so a getter is needed

PR #1841, opened 2025-06-10
[stable6] fix(PublicAuthPrompt): change default notice if identified

skjnldsv:
Pretty straightforward.
By moving to a prop approach, I forgot to make this dynamic by default


skjnldsv:
To fix after this is merged
https://github.com/nextcloud/server/pull/52537/files#r2138643592

skjnldsv:
> no main?

Will do after, since it was refactored

susnux (approved):

PR #1915, opened 2025-09-02
fix(guestName): limit length to 64 characters

max-nextcloud:
This is the same limit we apply to authenticated user display names.

The text app stores guest names in a 64 char database field.

Note that this is counting characters - not bytes.
So languages with multi-byte characters can still use 64 chars in name.

### DIFF
--- a/l10n/messages.pot
+++ b/l10n/messages.pot
@@ -100,6 +100,9 @@ msgstr ""
 msgid "Name"
 msgstr ""
 
+msgid "Names may be at most 64 characters long."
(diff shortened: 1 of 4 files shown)


max-nextcloud:
> Depending on the database used this will not work, it is only guaranteed to store 64 bytes then.

Could you point me to the database in question? According to https://docs.oracle.com/en/database/oracle/oracle-database/19/refrn/datatype-limits.html oracle does not seem to be an issue.

PR #1933, opened 2025-09-10
fix(FilePicker): make move primary and copy secondary

susnux:
* resolves https://github.com/nextcloud/server/issues/54969


jancborchardt:
Nice, thank you @susnux! :)

Issue #1526, opened 2024-12-14
Feature Request: Add Basic Yes/No Prompt Support


executed:
### Context:

The Nextcloud Calendar app currently includes a simple Yes/No confirmation dialog when performing actions like emptying the trashbin. However, this functionality relies on a deprecated, custom internal implementation of a confirm dialog. Developers are now encouraged to use the nextcloud-libraries/nextcloud-dialogs library, but this library does not currently provide built-in support for a basic Yes/No confirmation prompt.

This lack of support limits the ability to integrate simple confirmation dialogs into Nextcloud Calendar or other apps in a clean and maintainable way.

### Proposed Solution:

Enhance the nextcloud-libraries/nextcloud-dialogs library by adding support for basic Yes/No confirmation dialogs. This feature should:

Provide an API for simple confirmation prompts:

Allow developers to easily create a dialog with customizable text and "Yes"/"No" button labels.

Return a promise or similar mechanism to handle the user’s response (e.g., resolve with true for Yes, false for No).
### Example API:

nextcloudDialogs.confirm({
    message: "Are you sure you want to proceed?",
    title: "Confirm Action",
    confirmButton: "Yes",
    cancelButton: "No"
}).then(result => {
    if (result) {
        // User clicked "Yes"
    } else {
        // User clicked "No"
    }
});

### Example Use Case:

In the Nextcloud Calendar app, when an action like deleting an event or emptying the trashbin is performed, a Yes/No confirmation dialog can be shown using the new API. This dialog would replace the current deprecated implementation, providing a cleaner and more future-proof solution.Other than that I'm currently working on another confirmation dialog when resizing/dropping calendar event that is part of recurring series should ask user "Update future events too?".

susnux:
> This lack of support limits the ability to integrate simple confirmation dialogs into Nextcloud Calendar or other apps in a clean and maintainable way.

Whats the problem with just using the dialog builder?
https://nextcloud-libraries.github.io/nextcloud-dialogs/classes/index.DialogBuilder.html

PR #1986, opened 2025-09-30
feat(dialogs): allow to quickly show a dialog

susnux:
* resolves https://github.com/nextcloud-libraries/nextcloud-dialogs/issues/1526

Often its quite useful to just spin up a very simple dialog without need of using the builder.

(had this floating around for some time, so added tests now and created this PR :) ).


susnux:
rebased

PR #2033, opened 2025-10-16
feat(FilePicker): Improve pick and navigation logic to match user expectation

artonge:
- Set cursor as pointer on file rows
- Expose new `canPick` option
- Prevent navigation to non-readable nodes

| Before | After |
|--------|--------|
| image | Screenshot From 2025-10-17 11-17-19 |

- Needed by https://github.com/nextcloud/server/pull/55796

### DIFF
--- a/lib/composables/dav.spec.ts
+++ b/lib/composables/dav.spec.ts
@@ -64,7 +64,7 @@ describe('dav composable', () => {
 		vi.resetAllMocks()
 	})
 
-	it('Sets the inital state correctly', () => {
+	it('Sets the initial state correctly', () => {
 		const client = {
 			getDirectoryContents: vi.fn(() => ({ data: [] })),
 		}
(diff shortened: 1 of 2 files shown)


lib/components/FilePicker/FileListRow.vue
```diff
@@ -135,6 +140,11 @@ function handleKeyDown(event: KeyboardEvent) {
 		&--selected {
 			background-color: var(--color-background-dark);
 		}
+
+		&--not-pickable {
+			opacity: 0.5;
```
susnux: I think this is how we show it in most cases (also making colored previews less colorful):

```suggestion
			filter: saturate(0.7);
			opacity: 0.7;
```

lib/components/FilePicker/FileList.vue
```diff
@@ -67,7 +67,7 @@
 						:key="file.fileid || file.path"
 						:allow-pick-directory="allowPickDirectory"
 						:show-checkbox="multiselect"
-						:can-pick="multiselect || selectedFiles.length === 0 || selectedFiles.includes(file)"
+						:can-pick="(canPick === undefined || canPick(file)) && (multiselect || selectedFiles.length === 0 || selectedFiles.includes(file))"
```
susnux: Small performance improvement to not always execute the callback when not needed:


```suggestion
						:can-pick="(multiselect || selectedFiles.length === 0 || selectedFiles.includes(file)) && (canPick === undefined || canPick(file))"
```

So this way the cheap comparison is done on every file, but the potential expensive callback is only done when needed.

But this is theoretical - did not test it 😅

artonge: Not sure if it makes that much sense. But applied ;)

susnux (approved):

PR #2044, opened 2025-10-19
fix: compress translations in assets

susnux:
This reduces the bundle size within apps by about 200kb

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -17,8 +17,7 @@ const translations = readdirSync('./l10n')
 		// compress the translations
 		const translations = Object.entries(json.translations[''])
 			.filter(([key]) => key)
-			// eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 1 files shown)


vite.config.ts
```diff
@@ -7,18 +7,25 @@ import { readdirSync, readFileSync } from 'fs'
...
-		const locale = file.slice(0, -'.pot'.length)
+		const language = file.slice(0, -'.pot'.length)
 
 		const po = readFileSync(path)
 		const json = poParser.parse(po)
+		// compress the translations
+		const translations = Object.entries(json.translations[''])
+			.filter(([key]) => key)
+			// eslint-disable-next-line @typescript-eslint/no-unused-vars
+			.map(([_, value]) => value)
```
ShGKme:
```suggestion
			.map(([, value]) => value)
```

ShGKme (approved):

PR #2324, opened 2026-03-17
feat: accessible toast and drop toastify

skjnldsv:
Fix https://github.com/nextcloud-libraries/nextcloud-dialogs/issues/2254

:warning: Initially just a fix, afetr @susnux's comment, I decided it was a good thing to rewrite the implementation

### Drop `toastify-js`

Removes `toastify-js` (last third-party UI dep for toasts) and replaces it with a self-contained ToastNotification.vue. The public API is unchanged.

- Styling moved into the component using BEM, old SVG assets removed
- Close and undo buttons now use `NcButton`, spinner uses `NcLoadingIcon`
- Toasts slide in with a short fade+translate animation
- Height aligned to `--clickable-area-large`, positioned below the header
- 
### Accessibility

- Persistent `aria-live` regions (polite + assertive) are created once and reused, injecting a live region that already carries `aria-live` is unreliable in NVDA/JAWS, so they are pre-created before any toast is shown
- Announcements are prefixed with the toast type using a translatable `{message}` placeholder (`"Error: {message}"`) so the colour cue is also conveyed to screen readers (WCAG 1.4.1)
- The live region is synchronously cleared before each announcement so repeating the same message is always re-read
- `role="alert"` / `role="status"` set on each toast based on the resolved `ariaLive` level
- `NcLoadingIcon` is wrapped in `aria-hidden="true"` so the spinner doesn't pollute announcements

## Demo

https://github.com/user-attachments/assets/e5e25fdd-21ce-43fb-9c82-c1a82e0eb2c6


susnux (commented):
I wonder if we should completely remove the toasts dependency and write it from scratch as NcToast or similar to not have to monkey patch it all the time 😅

susnux (dismissed):
seems to work (did not test any screen reader!).
See my other comment but I think for now this is the best option.

szaimen (approved):
🐘

szaimen (approved):
🐘

susnux:
>  Having simultaneously two messages dispatched by one app and one call from another app seems unlikely, no ?

I can only imagine integrations like e.g. text in files or similar.

susnux (approved):

susnux (dismissed):

lib/components/ToastNotification.vue
```diff
@@ -0,0 +1,231 @@
...
+	/** ARIA role derived from the aria-live level */
+	role: 'alert' | 'status'
+	/** Optional click handler for the whole toast */
+	onClick?: () => void
+	/** Undo callback, only used when type is UNDO */
+	onUndo?: (event: MouseEvent) => void
+}>(), {
+	isHTML: false,
+	type: undefined,
+	close: true,
```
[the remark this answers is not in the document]
skjnldsv: Saw that eslint warning indeed. Sure thing!!

lib/toast.ts
```diff
@@ -1,13 +1,88 @@
...
+	const cached = level === 'assertive' ? _assertiveRegion : _politeRegion
+	if (cached && document.body.contains(cached)) {
+		return cached
+	}
+
+	const region = document.createElement('div')
+	region.setAttribute('aria-live', level)
+	region.setAttribute('aria-atomic', 'true')
+	region.setAttribute('aria-relevant', 'additions text')
+	region.style.cssText = VISUALLY_HIDDEN_STYLE
```
susnux: We do not need to re-declare the styles and can also just reuse the global classes:

```suggestion
	region.classList.add('hidden-visually')
```

susnux (commented):
I think one question we should solve would be if we want to keep HTML support, otherwise we could simplify this a lot.
We could still allow custom toast content by exposing the component from nextcloud-vue.

So in that case the component would have a `onBeforeCreate` that checks if the container is available or creates it if necessary.
And the `NcToast` component would roughly look like:
```vue
<template>
    <Teleport :to="toastContainerSelector">
        <div><!--- .... --->
            <slot><!-- ... --></slot>
        </div>
    </Teleport>
</template>
```

This would allow us to drop all of the custom "mounting" and announcement logic.
As then the announcement would also just teleport to the corresponding live region¹.


¹: This is currently a problem if you get 2 notifications quickly then the content is announced but directly cleared again and new notification is announced.
So teleporting into it would solve it because then its a list of active announcements.

GVodyanov:
Hey @skjnldsv, thanks for working on this. Is this still a PR you want to get in or should I make a new one, thanks!

skjnldsv:
@GVodyanov it should still go in at some point. But it might require a revive :)

skjnldsv:
Revived and addressed your comments Ferdinand, please feel free to have another look @susnux @GVodyanov

lib/toast.ts
```diff
@@ -1,13 +1,56 @@
...
+	if (typeof data === 'string') {
+		if (!isHTML) {
+			return data
+		}
+		// Parse in a standalone document (never attached to the page) so the
+		// markup is never live DOM, not just visually inert.
+		return getVisibleText(new DOMParser().parseFromString(data, 'text/html').body)
+	}
+	return getVisibleText(data)
+}
```
susnux: This should be placed below the exported consts and interfaces.
(you can also place them below exported functions if you like as they are functions they get hoisted on execution anyways and are defined when needed by other function)

lib/components/ToastContainer.vue
```diff
@@ -0,0 +1,189 @@
...
+<style module lang="scss">
+.toastContainer {
+	position: fixed;
+	// Default: no navigation sidebar
+	left: var(--body-container-margin);
+	bottom: var(--body-container-margin);
+	z-index: 10100;
+	display: flex;
+	flex-direction: column;
+	gap: 8px;
```
GVodyanov:
```suggestion
	gap: calc(var(--default-grid-baseline) * 2);
```

lib/components/ToastContainer.vue
```diff
@@ -0,0 +1,189 @@
...
+	align-items: flex-start;
+	// Individual toasts manage their own pointer-events
+	pointer-events: none;
+	padding: 8px;
```
GVodyanov:
```suggestion
	padding: calc(var(--default-grid-baseline) * 2);
```

GVodyanov:
A few comments from my end, I can add the code for them if you want @skjnldsv:
- I think currently the state warning/success/error etc is only represented by a color, and I think that isn't accessible. Some sort of invisible element for screen readers could be a solution
- One of the original complaints about this was that you couldn't pause the auto dismissal of notifications. A solution for this could be to not dismiss them if hovered/focused. We could have the toasts be focused when pressing <kbd>F6</kbd> like in Radix UI for example.
- The entrance animation should respect `@media (prefers-reduced-motion: reduce)`

lib/toast.ts
skjnldsv: Done

skjnldsv:
> * I think currently the state warning/success/error etc is only represented by a color, and I think that isn't accessible. Some sort of invisible element for screen readers could be a solution

We do announce it, this is the whole point of the aria live section.
https://github.com/nextcloud-libraries/nextcloud-dialogs/blob/d27342c04dade9e171f9ae642e8a1e039eeb4f2a/lib/toast.ts#L278-L291

> A few comments from my end, I can add the code for them if you want @skjnldsv:
> 
> * I think currently the state warning/success/error etc is only represented by a color, and I think that isn't accessible. Some sort of invisible element for screen readers could be a solution
> * One of the original complaints about this was that you couldn't pause the auto dismissal of notifications. A solution for this could be to not dismiss them if hovered/focused. We could have the toasts be focused when pressing F6 like in Radix UI for example.
> * The entrance animation should respect `@media (prefers-reduced-motion: reduce)`

Both of those points should be different PRs in my opinion :)

susnux (approved):

PR #2432, opened 2026-05-20
feat(ConflictPicker): allow to set overwrite flag

susnux:
Some operations like MOVE will not write into folder like the upload, but will just replace one folder with another.
Basically handle files and folders the same way.

In this case we need to indicate that the whole folder is to-be-replaced.

### DIFF
--- a/l10n/messages.pot
+++ b/l10n/messages.pot
@@ -208,6 +208,9 @@ msgstr ""
 msgid "When an incoming folder is selected, any conflicting files within it will also be overwritten."
 msgstr ""
 
+msgid "When an incoming folder is selected, any files within it will also be overwritten."
+msgstr ""
+
 msgid "When an incoming folder is selected, the content is written into the existing folder and a recursive conflict resolution is performed."
 msgstr ""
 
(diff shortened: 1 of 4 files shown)


Antreesy:
This is to be set from an app? Do you have an example? 
This is not a blocker, as don't see breaking changes here =) just to clarify

susnux:
> This is to be set from an app? Do you have an example?

Yes it needs to be set in the server "move file" action.

Given:
- a/
  - b/
    - Y
  - c.txt
- d/
  - b/
    - X
  - e.txt

in d/ you select all files and folders and click "move" and select target a/.
Then you get promoted because of the conflict "b".
You select the new one because the text said:

> When an incoming folder is selected, any **conflicting** files within it will also be overwritten.

But the result is:

- a/
  - b/
    - X
  - c.txt
  - e.txt

Because move overrides the folder. The old behavior was developed for this dialog with uploads in mind which write-into folders. But WebDAV MOVE overrides.

Antreesy (approved):
Looks fine then for the dialog change

PR #2539, opened 2026-07-30
feat: toast timeout configuration

kristian-zendato:
## Summary
Adds configurable timeouts for ordinary toast notifications.
- Exposes `getToastTimeout` and `setToastTimeout`
- Shares the configured value across bundled library copies via `window`
- Preserves fixed behavior for loading, undo, and permanent toasts

## 🤖 AI (if applicable)

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -23,6 +23,8 @@ export {
 } from './public-auth.ts'
 
 export {
+	getToastTimeout,
+	setToastTimeout,
 	showError,
 	showInfo,
 	showLoading,
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -23,8 +23,6 @@ export {
 } from './public-auth.ts'
 
 export {
-	getToastTimeout,
-	setToastTimeout,
 	showError,
 	showInfo,
 	showLoading,
--- a/lib/toast.ts
+++ b/lib/toast.ts
@@ -42,6 +42,48 @@ export const TOAST_DEFAULT_TIMEOUT = 7000
 export const TOAST_PERMANENT_TIMEOUT = -1
 
 /**
+ * Shared browser-global key for the configured toast timeout.
+ * Using `window` ensures separately bundled copies of this package share the same value.
+ */
+const GLOBAL_TOAST_TIMEOUT_KEY = '__nextcloud_dialogs_toast_timeout__'
+
+declare global {
+	interface Window {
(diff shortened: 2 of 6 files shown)


lib/index.ts
```diff
@@ -23,6 +23,8 @@ export {
 } from './public-auth.ts'
 
 export {
+	getToastTimeout,
+	setToastTimeout,
```
susnux: What would be the use case to expose this?
The values are only used internally by the toast implementation?

lib/toast.ts
susnux, line 1: I think this is only used internally so the value specified on server should be provided in the capabilities of `theming` and then used in the implementation here

lib/toast.ts
```diff
@@ -33,13 +35,51 @@ export enum ToastAriaLive {
...
+ * Falls back to {@link TOAST_DEFAULT_TIMEOUT} when unset or invalid.
+ */
+function getToastTimeout(): number {
+	try {
+		const timeout = (getCapabilities() as ThemingCapabilities)?.theming?.toastTimeout
+		if (typeof timeout === 'number' && isValidToastTimeout(timeout)) {
+			return timeout
+		}
+	} catch {
+		// Capabilities may be unavailable outside the browser (e.g. unit tests).
```
susnux: But then this should be rather mocked in test code then catch any possible exception here?

kristian-zendato: @susnux The exception is handled in test in here.
https://github.com/nextcloud-libraries/nextcloud-dialogs/pull/2539/changes#diff-6edceb110f8841a647b483a9687bf4acde4f737065e8fbe33c8037a87cc89da2R474-R477

susnux (approved):

lib/toast.ts
```diff
@@ -40,6 +41,37 @@ export const TOAST_DEFAULT_TIMEOUT = 7000
...
+ * Resolve the user-configured toast timeout from theming capabilities.
+ * Falls back to {@link TOAST_DEFAULT_TIMEOUT} when unset or invalid.
+ */
+function getToastTimeout(): number {
+	try {
+		const timeout = (getCapabilities() as ThemingCapabilities | undefined)?.theming?.toastTimeout
+		if (typeof timeout === 'number' && isValidToastTimeout(timeout)) {
+			return timeout
+		}
+	} catch (_error) {
```
susnux:
```suggestion
	} catch {
```

PR #2540, opened 2026-07-30
Simplify the file conflict dialog

jancborchardt:
Same as https://github.com/nextcloud-libraries/nextcloud-upload/pull/2151

Our current file conflict dialog (when you upload the same file again) immediately jumps into a very complex "pick files" view.

Based on a mockup by @kra-mo and some discussions.

The improvements:
- "Existing version" has been moved to the left and "New version" to the right. This is swapped from the previous layout as that was confusing – new things are usually on the right.
- The flow for 1 file was simplified by removing the checkboxes altogether, and just going for buttons "Keep both" and "Replace".
- For multiple files, the newer files are all preselected by default. That way people can click "Continue" directly.
- The wording is simplified
- To more easily compare what’s new or what changed between existing and new, whichever date is newer is bolded, and whichever size is larger is bolded.
- A little arrow inbetween each of the versions pointing right from the existing file to the new file helps with visual clarity.
- The icons of the cancel and skip buttons were removed to reduce visual noise.

We could probably simplify the wording even more for the multiple files case, but this PR is also quite big already. :) 

Before|After
-|-
before-single|single
before-multiple|multiple

## 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI

### DIFF
--- a/lib/components/ConflictPicker/ConflictPicker.vue
+++ b/lib/components/ConflictPicker/ConflictPicker.vue
@@ -7,7 +7,8 @@
 import type { INode } from '@nextcloud/files'
 import type { ConflictInput, ConflictResolutionResult } from '../../conflict-picker.ts'
 
-import { mdiArrowRight, mdiClose } from '@mdi/js'
+import { mdiArrowRight } from '@mdi/js'
+import { basename } from '@nextcloud/paths'
 import { computed, shallowRef, useTemplateRef } from 'vue'
 import NcButton from '@nextcloud/vue/components/NcButton'
 import NcCheckboxRadioSwitch from '@nextcloud/vue/components/NcCheckboxRadioSwitch'
@@ -25,9 +26,9 @@ const props = defineProps<{
 	container?: string | undefined
 
 	/**
-	 * Directory/context file name
+	 * Directory with the conflicts, its name is shown in the description
 	 */
-	dirname: string
+	dirname?: string
 
 	/**
 	 * The existing nodes (same names as the conflicts)
@@ -59,7 +60,10 @@ const blockedTitle = t('You need to select at least one version of each file to 
 const formElement = useTemplateRef('form')
 const conflictEntries = useTemplateRef('conflictEntry')
 
-const newSelected = shallowRef<ConflictInput[]>([])
+const isSingle = computed(() => props.incoming.length === 1)
+
+// Preselect the new files so the user can continue right away.
+const newSelected = shallowRef<ConflictInput[]>([...props.incoming])
 const oldSelected = shallowRef<ConflictInput[]>([])
 
 const isNoNewSelected = computed(() => newSelected.value.length === 0)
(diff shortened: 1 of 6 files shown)


lib/components/ConflictPicker/ConflictPickerCard.vue
```diff
@@ -0,0 +1,142 @@
+<!--
+  - SPDX-FileCopyrightText: 2023 Nextcloud GmbH and Nextcloud contributors
```
susnux:
```suggestion
  - SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
```

lib/components/ConflictPicker/ConflictPicker.vue
```diff
@@ -185,8 +223,10 @@ function onSubmit() {
 		<div :class="$style.pickerHeader">
 			<!-- Description -->
 			<p id="conflict-picker-description" :class="$style.pickerDescription">
-				{{ t('Which files do you want to keep?') }}<br>
-				{{ t('If you select both versions, the incoming file will have a number added to its name.') }}<br>
+				{{ descriptionParts.before }}<strong>{{ folderName }}</strong>{{ descriptionParts.after }}<br>
+				<template v-if="!isSingle">
+					{{ t('Files and folders not selected will be deleted. If both are chosen, they will be renamed.') }}<br>
```
susnux: This is a bit confusing, what does "will be deleted" mean?
In which case are files deleted?

lib/components/ConflictPicker/ConflictPickerEntry.vue
```diff
@@ -36,6 +36,12 @@ const props = defineProps<{
 	 * The new incoming node
 	 */
 	incoming: ConflictInput
+
+	/**
+	 * Single-file mode: render the two versions without checkboxes.
+	 * The parent then offers "Keep both"/"Replace" buttons instead.
+	 */
+	single?: boolean
```
susnux: in the other file its called `isSingle` maybe use it consistently (`isSingle`)

lib/components/ConflictPicker/ConflictPicker.vue
jancborchardt: @kra-mo since the wording is from you, do you have a better suggestion? :)

kra-mo: I guess you mean that it only applies to existing ones?

So that it should just be 

```suggestion
					{{ t('Existing files and folders will be deleted if not selected. If both are chosen, they will be renamed.') }}<br>
```

kra-mo (dismissed):
Nice, but some other things that we might as well change:
- "Incoming" I guess means the same as "new"? Probably good to keep the wording consistent.
- It would be nice to also change the padding for the buttons on the bottom to be consistent on the sides and the bottom.

susnux:
> "Incoming" I guess means the same as "new"? Probably good to keep the wording consistent.

Yes incoming are either
- Files to be uploaded (when used for upload conflicts)
- Files to be copied / moved (when used for that action)

jancborchardt:
@susnux all resolved. :) 

@kra-mo adjusted wording (to "new" as it’s much nicer and less technical) and bottom padding.

susnux (approved):

kra-mo (approved):

PR #2580, opened 2026-08-25
feat: cutomize file picker icon size

kristian-zendato:
Fixes: https://github.com/nextcloud-gmbh/customer-feature-requests/issues/1876

## Summary
Make the FilePicker list thumbnails configurable via the CSS custom property `--file-picker-preview-size` (default: `32px`), so instances can change size through theming without a public API.

## 🤖 AI (if applicable)

### DIFF
--- a/lib/components/FilePicker/FilePreview.vue
+++ b/lib/components/FilePicker/FilePreview.vue
@@ -7,17 +7,15 @@
 		:style="previewLoaded ? { backgroundImage: `url(${previewURL})` } : undefined"
 		:class="fileListIconStyles['file-picker__file-icon']">
 		<template v-if="!previewLoaded">
-			<IconFile v-if="isFile" :size="32" />
+			<IconFile v-if="isFile" />
 			<template v-else>
 				<NcIconSvgWrapper
 					v-if="folderDecorationIcon"
 					:class="fileListIconStyles['file-picker__file-icon-overlay']"
 					inline
 					:path="folderDecorationIcon"
-					:size="16" />
-				<IconFolder
-					:class="fileListIconStyles['file-picker__file-icon--primary']"
-					:size="32" />
+					size="auto" />
+				<IconFolder :class="fileListIconStyles['file-picker__file-icon--primary']" />
 			</template>
 		</template>
 	</div>
--- a/lib/composables/preview.ts
+++ b/lib/composables/preview.ts
@@ -13,9 +13,10 @@ import { preloadImage } from '../utils/imagePreload.ts'
 
 interface PreviewOptions {
 	/**
-	 * Size of the previews in px
+	 * Size of the previews in px.
+	 * Snapped to backend-pregenerated sizes (64 or 256); display size is controlled via CSS.
 	 *
-	 * @default 32
+	 * @default from `--file-picker-preview-size` (≤64 → 64, >64 → 256)
 	 */
 	size?: number
 	/**
@@ -39,7 +40,13 @@ interface PreviewOptions {
  * @param options Preview options
  */
 export function getPreviewURL(node: INode, options: PreviewOptions = {}) {
-	options = { size: 32, cropPreview: false, mimeFallback: true, ...options }
+	options = {
+		cropPreview: false,
+		mimeFallback: true,
+		...options,
+		// Only request pregenerated sizes so large folders do not flood the preview generator
+		size: toPregeneratedPreviewSize(options.size ?? getFilePickerPreviewRequestSize()),
+	}
 
 	try {
 		const previewUrl = node.attributes?.previewUrl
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -92,6 +92,19 @@ const filepicker = getFilePickerBuilder('Pick plain text files')
...
+`--file-picker-preview-size` (layout and preview request resolution stay in sync).
+
+To change it instance-wide (e.g. via theming / custom CSS), without exposing a public FilePicker API:
+
+```css
+:root {
+	--file-picker-preview-size: 64px;
+}
+```
+
```
susnux: This is nothing we want to be used, I would not document this here.
For consistency every app in Nextcloud should use the default.

lib/components/FilePicker/FilePreview.vue
```diff
@@ -52,10 +52,16 @@ const props = defineProps<{
 // workaround for vue2.7 bug, can be removed with vue3
 const fileListIconStyles = ref(fileListIconStylesModule)
 
+const previewSize = getFilePickerPreviewSize()
```
susnux: I do not think this is needed, all of this can be done with pure CSS without the need to get the value from CSS -> JS -> CSS.

lib/composables/preview.ts
```diff
@@ -11,11 +11,41 @@ import { generateUrl } from '@nextcloud/router'
...
+const FILE_PICKER_PREVIEW_SIZE_DEFAULT = 32
+
+/**
+ * Resolve the FilePicker preview size from `:root` CSS (or the default).
+ * Used only for preview *request* resolution so fetched images match display size.
+ */
+function getFilePickerPreviewSize(): number {
+	if (typeof window === 'undefined' || typeof getComputedStyle === 'undefined') {
+		return FILE_PICKER_PREVIEW_SIZE_DEFAULT
+	}
```
susnux: This library is only intended for window usage so this is only needed for tests without DOM.
So move this to those tests instead and remove from production

lib/composables/preview.ts
```diff
@@ -11,11 +11,41 @@ import { generateUrl } from '@nextcloud/router'
...
+
+	const raw = getComputedStyle(document.documentElement)
+		.getPropertyValue(FILE_PICKER_PREVIEW_SIZE_VAR)
+		.trim()
+	const parsed = Number.parseFloat(raw)
+	if (Number.isFinite(parsed) && parsed > 0) {
+		return Math.round(parsed)
+	}
+
+	return FILE_PICKER_PREVIEW_SIZE_DEFAULT
```
susnux: We should cache this because this function can be called thousands of times.
e.g. put a `let previewSize: number` in the module and then check if its `=== undefined` and if not return it otherwise set it with this logic here

lib/composables/preview.ts
```diff
@@ -39,7 +69,13 @@ interface PreviewOptions {
  * @param options Preview options
  */
 export function getPreviewURL(node: INode, options: PreviewOptions = {}) {
-	options = { size: 32, cropPreview: false, mimeFallback: true, ...options }
+	options = {
+		cropPreview: false,
+		mimeFallback: true,
+		...options,
+		// Keep request resolution in sync with CSS display size (avoids blurry upscales)
+		size: options.size ?? getFilePickerPreviewSize(),
```
susnux: We can do a performance improvement here and always query `64px`  by default.
Because this is the size pregenerated on the backend, scaling can then be done with CSS.
In general we have two pregenerated (cheap) sizes:
- 64px
- 256px

kristian-zendato: But if we always query `64px`, the preview icon would be blurry for bigger icons size(e.g. `128px` or `256px`).
image

susnux: sure but the use case is is basically our default of ~34px (?) and 10px.
This method would be called hundreds or thousands of times for big folders.

So the value should be at least only calculated once on first usage (css should not change anymore) and then clamped size <= 64px -> use 64px, size > 64px -> use 256px.
Otherwise this will generate too many previews causing troubles on big instances

lib/composables/preview.ts
```diff
@@ -11,11 +11,60 @@ import { generateUrl } from '@nextcloud/router'
...
+		.getPropertyValue(FILE_PICKER_PREVIEW_SIZE_VAR)
+		.trim()
+	const parsed = Number.parseFloat(raw)
+	const displaySize = Number.isFinite(parsed) && parsed > 0
+		? Math.round(parsed)
+		: PREGENERATED_PREVIEW_SIZE_SMALL
+
+	previewRequestSize = toPregeneratedPreviewSize(displaySize)
+	return previewRequestSize
+}
```
susnux: Makes sense, but for readability I would move all this below the exported stuff, as this is just helper methods (and `function` is hoisted in JS any ways).

susnux (approved):

PR #2611, opened 2026-09-12
fix(dav): keep the latest FilePicker destination

Daddyoftwins1:
## Problem

While moving a folder in Nextcloud, I selected Photos as the destination, but the picker returned the folder I was already in and displayed the message that the folder was already in that directory.

The problem can happen when directory requests finish out of order. Opening another folder starts a newer DAV listing request, but cancellation does not guarantee that the older request stops immediately. If the older response finishes later, it can replace the active folder information even though the path and button still show the newer destination.

This leaves the displayed destination and the folder returned by the picker out of sync.

## Changes

This change:

* ensures that only the latest DAV listing request can update the current folder, file list and loading state
* prevents an older response from replacing a newer destination
* clears stale destination information while a replacement folder is loading or when its request fails
* prevents directory confirmation until the loaded folder matches the path displayed by the picker
* adds regression tests for delayed responses, cancellation ordering, failed requests, view changes and destination confirmation

Related to #2511, which covers the same race-condition area.

## Testing

The following focused checks passed:

* 8 Vitest tests covering the changed DAV and FilePicker behavior
* ESLint on all four changed files
* `git diff --check`

The complete repository test suite, a production build and fresh live-instance testing were not performed for this contribution branch.

## AI assistance

Development and automated testing of this contribution were assisted by OpenAI Codex using GPT-5.


susnux:
Thank you for your pull request, could you please rebase your branch?

Daddyoftwins1:
> Thank you for your pull request, could you please rebase your branch?

Thanks! I rebased the branch onto the latest main. The focused test suite now passes 27 tests, and ESLint and whitespace checks also pass.

